PR #2, opened 2014-02-05
[FIXED JENKINS-21374] Allow disabling ROLE_ creation

daniel-beck:
Adds another option to disable creation of `ROLE_` roles for those of us who've never used them, and never will.

Advantages:
- No role duplication (reduces roles/groups list by half when accessing a user's page)
- No user confusion

Another option would be to add another system property if that's preferred over another option.


oleg-nenashev:
:+1: , but needs a re-merging with the master

daniel-beck:
Apparently superseded by @stephenc's https://github.com/jenkinsci/ldap-plugin/pull/18/commits/3d556a31e7fdcfaaf23e2ba0f4d8e0a61d5da201

PR #3, opened 2015-03-04
this change allows the change of the memberOf attribute

ritchieGitHub:
Some organisations have other attributes than memberOf that specifies the groups.
With this pull request it is changeable.


oleg-nenashev:
Thanks for catching the issue. 
Do you really need a setter method? If no, I would just fix the jelly file.

oleg-nenashev:
Somebody also should create an issue in JIRA

ritchieGitHub:
Without the setter the script failed in my system, if the current version can do without then it is not nessesary anymore.

oleg-nenashev:
@ritchieGitHub 
The constructor is correct, so it should work. I'll try to check it on this week.

oleg-nenashev:
I confirm the code works for me without a setter. I can just pick https://github.com/entis-group/ldap-plugin/commit/1b87f243a70e356b40b0c3f52a039acbb9d48b61 to the master branch

ritchieGitHub:
thanks!

KostyaSha:
right, should be `attributeName` in tact with constructor

JesperTerkelsen:
Bump - just ran into this
Why was this never merged ?

rdpa:
Bump again. This is giving us trouble right now.

andresrc:
Included in https://github.com/jenkinsci/ldap-plugin/pull/10

PR #4, opened 2015-07-16
Enable FindBugs in the plugin

oleg-nenashev:
- [x] Enable FindBugs
- [x] Cleanup the most important issues

@reviewbybees

### DIFF
--- a/src/main/java/hudson/security/LDAPSecurityRealm.java
+++ b/src/main/java/hudson/security/LDAPSecurityRealm.java
@@ -24,6 +24,7 @@
  */
 package hudson.security;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import groovy.lang.Binding;
 import hudson.DescriptorExtensionList;
 import hudson.Extension;
@@ -64,6 +65,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import javax.annotation.Nonnull;
 import javax.naming.Context;
 import javax.naming.NamingException;
 import javax.naming.directory.Attribute;
@@ -98,6 +100,8 @@ import org.acegisecurity.userdetails.ldap.LdapUserDetailsImpl;
 import org.apache.commons.collections.map.LRUMap;
 import org.apache.commons.io.input.AutoCloseInputStream;
 import org.apache.commons.lang.StringUtils;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
 import org.springframework.dao.DataAccessException;
@@ -239,6 +243,8 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
      * LDAP server name(s) separated by spaces, optionally with TCP port number, like "ldap.acme.org"
      * or "ldap.acme.org:389" and/or with protcol, like "ldap://ldap.acme.org".
      */
+    @SuppressFBWarnings(value = "UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD", 
+        justification = "This public field is exposed to the plugin's API")
     public final String server;
 
     /**
@@ -246,12 +252,16 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
      *
      * How do I infer this?
      */
+    @SuppressFBWarnings(value = "UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD", 
+        justification = "This public field is exposed to the plugin's API")
     public final String rootDN;
 
     /**
      * Allow the rootDN to be inferred? Default is false.
      * If true, allow rootDN to be blank.
      */
+    @SuppressFBWarnings(value = "UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD", 
+        justification = "This public field is exposed to the plugin's API")
     public final boolean inhibitInferRootDN;
 
     /**
@@ -260,6 +270,8 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
      *
      * Something like "ou=people" but can be empty.
      */
+    @SuppressFBWarnings(value = "UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD", 
+        justification = "This public field is exposed to the plugin's API")
     public final String userSearchBase;
 
     /**
@@ -269,6 +281,8 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
      *
      * @see FilterBasedLdapUserSearch
      */
+    @SuppressFBWarnings(value = "UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD", 
+        justification = "This public field is exposed to the plugin's API")
     public final String userSearch;
     
     /**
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
@reviewbybees
Ready for the review

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -230,6 +232,8 @@
  * @author Kohsuke Kawaguchi
  * @since 1.166
  */
+@SuppressFBWarnings(value = "UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD", 
+        justification = "This class exposes public fields as a part of its API")
```
stephenc: I would favour this on all existing public fields so that nobody can add more fields that are public. In general everywhere there is a public field as part of the API it has been a mistake

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -628,17 +632,28 @@ public String getMailAddressAttributeName() {
...
     }
 
-    public SecurityComponents createSecurityComponents() {
+    /**
+     * Creates security components.
+     * @return Created {@link SecurityComponents}
+     * @throws Error Execution error
+     */
+    @Override
+    public @Nonnull SecurityComponents createSecurityComponents() {
```
stephenc: When annotating the method, please annotate the method. Just because some annotations have semantic meaning on the return type does not mean you should pretend you are annotating the return type. You are annotating the method.

stephenc:
:bee: on the over-zealous scope of the public fields API and the incorrect placement of the annotation on a method.

src/main/java/hudson/security/LDAPSecurityRealm.java
oleg-nenashev: Yes, makes sense

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -628,17 +654,28 @@ public String getMailAddressAttributeName() {
...
+     */
+    @Override @Nonnull
     public SecurityComponents createSecurityComponents() {
         Binding binding = new Binding();
         binding.setVariable("instance", this);
 
-        BeanBuilder builder = new BeanBuilder(Jenkins.getInstance().pluginManager.uberClassLoader);
+        final Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins == null) {
+            throw new Error("Jenkins instance is not ready");
```
jtnord: NIT: I prefer throwing an `IllegalStateException` which would allow for code cleanup using `Jenkins.getActiveInstance()` in the future.

oleg-nenashev: Makes sense. I've used Error because there were other usages in the code below

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1109,6 +1159,8 @@ private static String addPrefix(String server) {
      * See http://msdn.microsoft.com/en-us/library/aa746475(VS.85).aspx for the syntax by example.
      * WANTED: The specification of the syntax.
      */
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", 
+            justification = "Filter may be changed in the runtime by scripts if required")
```
jtnord: This is the DEFAULT if I'm not mistaken so should never be changed by scripts and can be made final.

src/main/java/hudson/security/LDAPSecurityRealm.java
jtnord: "Urgh" on the existing `throw new Error(...)` - anyway - I'm not voting bug on this :-)

oleg-nenashev: I'll fix it in a separate pull request

src/main/java/hudson/security/LDAPSecurityRealm.java
oleg-nenashev: BTW it's a common practice to keep such variables modifiable. See Jenkins core

jtnord: yes - some would say it is a common **bad** practice.
changing this at runtime is likely to have zero affect on a running system.

stephenc: I think this should just be made `final`. The reason it was non-final in the past was because the value was not exposed in the UI. A while back we switched to have the filter configurable via the UI, so no need to have this value live-modifiable any more.

oleg-nenashev: @stephenc 
I doubt it can be changed without breaking the backward compatibility. IMO it does not worth it

stephenc: If somebody is trying to write to this field from the groovy console (which is the only place you would try now) then they can stop. If they were trying to write to it from `init.groovy`, then they should stop. Make the field final to help them

oleg-nenashev: If somebody is trying to write to this field from the Java code then something will blow up... Even if it is a crazy thing to do

oleg-nenashev: Anyway, it should be done by a separate pull request, because it is not a "code" polishing. @stephenc , do you agree with merging this PR as is? If yes, I'll create an issue for this field

stephenc: I do not agree with the current `@SuppressFBWarnings`. Either fix the justification and add the TODO to make the field final or just remove the `@SuppressFBWarnings` and make the field final

oleg-nenashev:
@stephenc @jtnord 
I've addressed the comments regarding LDAPSecurityRealm#GROUP_SEARCH. The field is also restricted now.

jtnord:
:bee:

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1109,6 +1162,9 @@ private static String addPrefix(String server) {
      * See http://msdn.microsoft.com/en-us/library/aa746475(VS.85).aspx for the syntax by example.
      * WANTED: The specification of the syntax.
      */
+    @Restricted(NoExternalUse.class)
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", 
+            justification = "Filter may be changed in the runtime by plugins, API backward compatibility")
```
stephenc: what plugins do that? Seriously just make the field `final`

oleg-nenashev: Sorry, it would be a formal backward compatibility issue. I won't do it without enough justification.

stephenc: Fine well we can merge #5 first then

stephenc: Or to put it another way, I strongly object to adding incorrectly labelled `@SuppressFBWarnings`

oleg-nenashev: In #5 I see a justification at least. If we have somebody wearing the maintainer's hat, he should just make a decision on that case.

stephenc: That would be me wearing the maintainer's hat... which is why I have been telling you the field should be final

stephenc: To put it another way, 10 of the last 11 releases were by me

amuniz:
:bee:

jtnord:
:bee:

stephenc:
:bee:

oleg-nenashev:
@reviewbybees done

src/main/java/hudson/security/LDAPSecurityRealm.java
KostyaSha: :-1: core uses IllegalState, using Error inconsistent

oleg-nenashev: @KostyaSha 
As we discussed above, it's a topic for another PR. It's in my TODO list

PR #5, opened 2015-07-27
Make GROUP_SEARCH final as run-time mutability has been obsoleted by …

stephenc:
…other changes

This field was runtime mutable only because prior to the group search being exposed via the UI it was the only way to modify the search.

When the group search was exposed in the UI this field's behaviour was changed from being the actual value used to being the initial default value and as such there is no need for it to be run-time modifiable.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/security/LDAPSecurityRealm.java
+++ b/src/main/java/hudson/security/LDAPSecurityRealm.java
@@ -1109,7 +1109,7 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
      * See http://msdn.microsoft.com/en-us/library/aa746475(VS.85).aspx for the syntax by example.
      * WANTED: The specification of the syntax.
      */
-    public static String GROUP_SEARCH = System.getProperty(LDAPSecurityRealm.class.getName()+".groupSearch",
+    public static final String GROUP_SEARCH = System.getProperty(LDAPSecurityRealm.class.getName()+".groupSearch",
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Needs a binary compatibility definition in pom.xml at least

stephenc:
> Needs a binary compatibility definition in pom.xml at least

If we followed that principle then this plugin would be heading towards version `9.0.0`

Semver is too naïve to follow, e.g. see some of my [thoughts on the matter](http://javaadventure.blogspot.ie/2014/09/version-numbers.html).

In the case of Jenkins plugins, the majority convention is to bump the major version number if you are making a backwards incompatible serialized file format change. We are not making such a change here.

oleg-nenashev:
@stephenc 
Sorry, I've improperly referenced the approach. My idea was to print a warning on the UC by hooking on <code>UpdateSite.Plugin.isCompatibleWithInstalledVersion()</code>. I've though it comes from Maven POM, but seems, I'm wrong

If such approach is to complex, a notice on Wiki should be enough. I'm OK with the change after the justification. :bee:

stephenc:
`UpdateSite.Plugin.isCompatibleWithInstalledVersion()` does come from the Maven pom **but** it is for use where the on-disk file format has changed. Other compat contracts are out of scope from the warning message that is triggered with that flag

varmenise:
:bee:

stephenc:
@reviewbybees done

stephenc:
(as plugin maintainer) :+1:

PR #6, opened 2015-08-19
Throw IllegalStateException instead of the Error

oleg-nenashev:
Follow-up to #4 ...
The Javadoc has not been released yet, so the change should not break anything on the user side.

@reviewbybees @KostyaSha


amuniz:
:bee:

KostyaSha:
:+1:

oleg-nenashev:
@reviewbybees done

PR #7, opened 2016-03-03
[JENKINS-33300] Upgrade to parent pom 2.3

armfergom:
@reviewbybees

* Upgraded to parent pom 2.3
* Fixed javadoc issue

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.566</version>
+    <version>2.3</version>
   </parent>
 
   <artifactId>ldap</artifactId>
@@ -23,9 +23,8 @@
   </licenses>
 
   <properties>
-    <maven-hpi-plugin.version>1.96</maven-hpi-plugin.version>
-    <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
-    <findbugs.failOnError>true</findbugs.failOnError>
+    <jenkins.version>1.566</jenkins.version>
+    <java.level>6</java.level>
   </properties>
 
   <scm>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -53,78 +52,27 @@
...
     <plugins>
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>animal-sniffer-maven-plugin</artifactId>
-        <!-- details defined in the parent POM -->
-      </plugin>
-      <plugin>
         <groupId>org.codehaus.gmaven</groupId>
         <artifactId>gmaven-plugin</artifactId>
         <executions>
```
recena: Really needed?

armfergom: Yes, I think so. If removed, the version used will be a greater one and the plugin stops building due to an API incompatibility. Specifically, this one:

```
 [ERROR] Failed to execute goal org.codehaus.gmaven:gmaven-plugin:1.5-jenkins-3:testCompile (test-in-groovy) on project ldap: Execution test-in-groovy of goal org.codehaus.gmaven:gmaven-plugin:1.5-jenkins-3:testCompile failed: An API incompatibility was encountered while executing org.codehaus.gmaven:gmaven-plugin:1.5-jenkins-3:testCompile: java.lang.NoSuchMethodError: org.codehaus.groovy.ast.ModuleNode.getStarImports()Ljava/util/List;
```

armfergom:
One more commit removing the `compatibleSinceVersion`configuration parameter for the maven-hpi-plugin. Not needed since parent pom version 1.321 as stated [here](https://wiki.jenkins-ci.org/display/JENKINS/Marking+a+new+plugin+version+as+incompatible+with+older+versions)

pom.xml
```diff
@@ -53,103 +52,33 @@
...
-        <version>2.5</version>
       </plugin>
       <plugin>
         <groupId>org.jenkins-ci.tools</groupId>
         <artifactId>maven-hpi-plugin</artifactId>
-        <version>${maven-hpi-plugin.version}</version>
         <configuration>
-          <compatibleSinceVersion>1.9</compatibleSinceVersion>
           <systemProperties>
             <hudson.bundled.plugins>${basedir}/work/plugins/ldap.hpl</hudson.bundled.plugins>
```
andresrc: We should see if this is still needed, and if not remove the whole plugin entity.

armfergom: When the plugin is bundled, this is needed so that the HPI plugin uses the plugin being developed and not the one bundled. Same happens for example with the Subversion plugin.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1291,7 +1291,7 @@ public String getValue() {
 
             @Override
             public String getDisplayName() {
-                return null;
+                return "";
```
amuniz: This seems to be not related to the parent pom upgrade. What's different on returning `null` or empty string? Is it fixing a bug?

armfergom: With the new parent pom, findbugs is integrated. With the current version of the core (1.566), the plugin builds ok and findbugs does not find issues.

However, if this is run against the latest core version, findbugs starts complaining because the method getDisplayName() in the superclass is annotated with @Nonnull. It is important to avoid this, so that the plugin compatibility test works correctly.

andresrc: @amuniz We are trying to get a clean  `mvn clean verify -Djenkins.version=last_lts` build as well when upgrading to the parent POM, to anticipate possible PCT issues.

andresrc:
:bee:

armfergom:
@reviewbybees done

abayer:
:bee:, merging.

PR #9, opened 2016-06-15
[JENKINS-8152] Encode rootDN in provider URL.

andresrc:
See [JENKINS-8152](https://issues.jenkins-ci.org/browse/JENKINS-8152)

The rootDN must be URI-encoded when included in the provider URL.

@reviewbybees

### DIFF
--- a/src/test/java/hudson/security/LDAPSecurityRealm_Test.java
+++ b/src/test/java/hudson/security/LDAPSecurityRealm_Test.java
@@ -32,7 +32,9 @@ import static org.junit.Assert.*;
 import jenkins.security.plugins.ldap.FromGroupSearchLDAPGroupMembershipStrategy;
 import org.junit.Rule;
 import org.junit.Test;
+import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.WithoutJenkins;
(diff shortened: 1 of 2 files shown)


recampbell:
@jtnord perhaps you can take a look at this 2mo old PR?

src/test/java/hudson/security/LDAPSecurityRealm_Test.java
```diff
@@ -68,4 +70,18 @@ private void check() {
         assertEquals("mAAN", sr.getMailAddressAttributeName());
     }
 
+    @Issue("JENKINS-8152")
+    @WithoutJenkins
+    @Test public void providerUrl() throws Exception {
+        assertEquals("ldap://example.com/", LDAPSecurityRealm.toProviderUrl("example.com", null));
+        assertEquals("ldap://example.com/", LDAPSecurityRealm.toProviderUrl("example.com", ""));
+        assertEquals("ldap://example.com/", LDAPSecurityRealm.toProviderUrl("example.com", "   "));
```
jtnord: rootDN is trimmed in the constrictor - so this is testing behaviour that should never occur.

jtnord:
:bee: seems reasonable - concern would be over people that have manually escaped things already, as an upgrade would break their ability to login.

recampbell:
@reviewbybees  done.

PR #11, opened 2017-01-31
[JENKINS-38124] using FromUser groupmembership when is configured

escoem:
IMO this is a security issue because users may have different permissions than is expected as you could easily see at JIRA comment https://issues.jenkins-ci.org/browse/JENKINS-38124?focusedCommentId=285709&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-285709 cc. @daniel-beck we can discuss that if you see it appropriate.

That PR fix the issue but could be considered as WIP if the maintainer prefers. cc, @andresrc 

junit test hasn't been added because there wasn't before but a test case is described at https://issues.jenkins-ci.org/browse/JENKINS-38124?focusedCommentId=285709&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-285709

cc. @aheritier @fbelzunc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,6 +74,7 @@
         <groupId>org.jenkins-ci.tools</groupId>
         <artifactId>maven-hpi-plugin</artifactId>
         <configuration>
+          <compatibleSinceVersion>1.15</compatibleSinceVersion>
           <systemProperties>
             <hudson.bundled.plugins>${basedir}/work/plugins/ldap.hpl</hudson.bundled.plugins>
           </systemProperties>
--- a/src/main/java/hudson/security/LDAPSecurityRealm.java
+++ b/src/main/java/hudson/security/LDAPSecurityRealm.java
@@ -24,6 +24,7 @@
  */
 package hudson.security;
 
+import com.google.common.collect.Sets;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import groovy.lang.Binding;
 import hudson.DescriptorExtensionList;
@@ -79,6 +80,7 @@ import javax.naming.directory.InitialDirContext;
 import jenkins.model.IdStrategy;
 import jenkins.model.Jenkins;
 import jenkins.security.plugins.ldap.FromGroupSearchLDAPGroupMembershipStrategy;
+import jenkins.security.plugins.ldap.FromUserRecordLDAPGroupMembershipStrategy;
 import jenkins.security.plugins.ldap.LDAPGroupMembershipStrategy;
 import org.acegisecurity.AcegiSecurityException;
 import org.acegisecurity.Authentication;
@@ -994,17 +996,24 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
         // Make these available (private in parent class and no get methods!)
         String rolePrefix = "ROLE_";
         boolean convertToUpperCase = true;
+        LDAPGroupMembershipStrategy groupMembershipStrategy;
+        LdapUserSearch ldapSearch;
 
-        public AuthoritiesPopulatorImpl(InitialDirContextFactory initialDirContextFactory, String groupSearchBase) {
+        public AuthoritiesPopulatorImpl(InitialDirContextFactory initialDirContextFactory, String groupSearchBase, LDAPGroupMembershipStrategy groupMembershipStrategy) {
             super(initialDirContextFactory, fixNull(groupSearchBase));
 
             super.setRolePrefix("");
             super.setConvertToUpperCase(false);
+            this.groupMembershipStrategy = groupMembershipStrategy;
         }
 
         @Override
         protected Set getAdditionalRoles(LdapUserDetails ldapUser) {
             return Collections.singleton(AUTHENTICATED_AUTHORITY);
+        }
+
+        public void setLdapSearch(LdapUserSearch ldapSearch) {
+            this.ldapSearch = ldapSearch;
         }
 
         @Override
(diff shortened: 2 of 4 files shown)


src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -994,19 +996,26 @@ public String findMailAddressFor(User u) {
         // Make these available (private in parent class and no get methods!)
         String rolePrefix = "ROLE_";
         boolean convertToUpperCase = true;
+        LDAPGroupMembershipStrategy groupMembershipStrategy;
+        LdapUserSearch ldapSearch;
 
-        public AuthoritiesPopulatorImpl(InitialDirContextFactory initialDirContextFactory, String groupSearchBase) {
+        public AuthoritiesPopulatorImpl(InitialDirContextFactory initialDirContextFactory, String groupSearchBase, LDAPGroupMembershipStrategy groupMembershipStrategy) {
```
andresrc: Binary incompatibility.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1028,21 +1037,43 @@ public void setConvertToUpperCase(boolean convertToUpperCase) {
          */
         @Override
         public Set getGroupMembershipRoles(String userDn, String username) {
-            Set<GrantedAuthority> names = super.getGroupMembershipRoles(userDn,username);
+            if (groupMembershipStrategy instanceof FromUserRecordLDAPGroupMembershipStrategy) {
```
andresrc: Maybe better to encapsulate this logic in the strategy.

src/main/java/hudson/security/LDAPSecurityRealm.java
escoem: not really because it's used internally from LDAPBindSecurityRealm.groovy

src/main/java/hudson/security/LDAPSecurityRealm.java
escoem: No, it's required at AuthoritiesPopulatorImpl and also outside (loaduserfromid)

aheritier:
LGTM +1

src/main/resources/hudson/security/LDAPBindSecurityRealm.groovy
```diff
@@ -65,10 +65,11 @@ bindAuthenticator(BindAuthenticator2,initialDirContextFactory) {
     userSearch = ldapUserSearch;
 }
 
-authoritiesPopulator(AuthoritiesPopulatorImpl, initialDirContextFactory, instance.groupSearchBase) {
+authoritiesPopulator(AuthoritiesPopulatorImpl, initialDirContextFactory, instance.groupSearchBase, instance.groupMembershipStrategy) {
     // see DefaultLdapAuthoritiesPopulator for other possible configurations
     searchSubtree = true;
     groupSearchFilter = "(| (member={0}) (uniqueMember={0}) (memberUid={1}))";
+    ldapSearch = ldapUserSearch;
```
marvinthepa: Do I understand it correctly that this file can also be overridden by the user? See https://wiki.jenkins-ci.org/display/JENKINS/LDAP+Plugin#LDAPPlugin-Troubleshooting.

In that case, does this break if users have overridden this file without the additional constructor argument?

escoem: You are right... thanks, I need to review and maybe use a setter. But maybe an update of those files should be required :(

escoem: @kohsuke @stephenc can you help us with this file please? does it still make sense to be overwritten? IIUC, you can only setup timeout and/or groupsearch filter. Timeouts can be setted as java parameters and groupsearch query can be setup through UI. Or just removed from https://github.com/jenkinsci/ldap-plugin/blob/master/src/main/java/hudson/security/LDAPSecurityRealm.java#L698-L701

src/main/java/hudson/security/LDAPSecurityRealm.java
stephenc: ahhh but what about an installation that has a custom `LDAPBindSecurityRealm.groovy` that they haven't touched in years... with this change you will break their instance and they cannot login to figure out why

src/main/resources/hudson/security/LDAPBindSecurityRealm.groovy
stephenc: > does it still make sense to be overwritten? 

No, but there are instances that did override it back when that was the only way, and this plugin has to stay working for those people... OTOH you can assume that they do not have fancy configurations in that case

escoem:
@stephenc does my last changes address your comments?

escoem:
hey @stephenc are you there?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1028,21 +1040,66 @@ public void setConvertToUpperCase(boolean convertToUpperCase) {
          */
         @Override
         public Set getGroupMembershipRoles(String userDn, String username) {
-            Set<GrantedAuthority> names = super.getGroupMembershipRoles(userDn,username);
+            if (ldapSearch == null) {
+                LOGGER.log(Level.WARNING, "[JENKINS-38124] Please update $JENKINS_HOME/LDAPBindSecurityRealm.groovy file to fix it.");
```
stephenc: Create a sub-page on the LDAP plugin wiki page with details of how somebody is supposed to fix this issue and include a link to that sub-page

pom.xml
```diff
@@ -74,6 +74,7 @@
         <groupId>org.jenkins-ci.tools</groupId>
         <artifactId>maven-hpi-plugin</artifactId>
         <configuration>
+          <compatibleSinceVersion>1.15</compatibleSinceVersion>
```
stephenc: Is this true? If I upgrade to 1.15 I can no longer downgrade to 1.14 without restoring the configuration from a backup?

daniel-beck: We only have this one flag, so it's often applied "creatively" and not strictly based on what it says. In this case, AFAIUI, people may lose permissions, so we really want admins to read the changelog beforehand.

stephenc: Yes, I would prefer at a minimum to have a comment explaining why beside this entry in the Pom so that future explorers can understand the reasoning

escoem:
Hey, it seems that this PR has no sense currently so I'm going to close it. Thanks for your reviews

PR #13, opened 2017-03-08
[JENKINS-28474] workaround chrome autocomplete issue

Twiknight:
fix for https://issues.jenkins-ci.org/browse/JENKINS-28474.
Now `autocomplete="off"` does not work in some case.
Use `autocomplete="new-password"` will require a manually inputed new password

See:
http://stackoverflow.com/questions/15738259/disabling-chrome-autofill
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete

### DIFF
--- a/src/main/resources/hudson/security/LDAPSecurityRealm/config.jelly
+++ b/src/main/resources/hudson/security/LDAPSecurityRealm/config.jelly
@@ -65,7 +65,7 @@ THE SOFTWARE.
             <f:textbox autocomplete="off"/>
         </f:entry>
         <f:entry field="managerPasswordSecret" title="${%Manager Password}">
-            <f:password autocomplete="off"/>
+            <f:password autocomplete="new-password"/>
         </f:entry>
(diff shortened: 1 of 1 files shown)


Twiknight:
This problem is fatal.
If you're autofilled with chrome, you may get locked outside Jenkins before you can get the system operation team to fix the config file from server side.

fabriziocucci:
I agree with @Twiknight, this problem is quite serious.

I've experienced the issue of being locked out from Jenkins (as reported [here](https://github.com/jenkinsci/ldap-plugin/pull/13)).

Why there is no one even assigned to this?

zbynek:
Didn't @daniel-beck fix this in https://github.com/jenkinsci/jenkins/commit/90e693b13bfb32a09e2d5ebe4f9b4a0da7d51979 ?

daniel-beck:
That would be my expectation. I suggest we close this and the referenced issue saying we believe it's resolved since 2.205, and anyone who still experiences the problem should file a new issue.

PR #14, opened 2017-03-24
Update baseline version of Jenkins to 1.625.3

stephenc:
Also add a Jenkinsfile

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.21</version>
+    <version>2.25</version>
   </parent>
 
   <artifactId>ldap</artifactId>
@@ -23,8 +23,8 @@
   </licenses>
 
   <properties>
-    <jenkins.version>1.566</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins.version>1.625.3</jenkins.version>
+    <java.level>7</java.level>
   </properties>
 
   <scm>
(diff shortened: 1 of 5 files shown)


stephenc:
@rsandell I found some todos that could be fixed now BTW

pom.xml
```diff
@@ -23,8 +23,8 @@
   </licenses>
 
   <properties>
-    <jenkins.version>1.566</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins.version>1.625.3</jenkins.version>
+    <java.level>7</java.level>
```
jglick: May omit as it is the default.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -467,16 +467,7 @@ public LDAPSecurityRealm(String server, String rootDN, String userSearchBase, St
         this(server, rootDN, userSearchBase, userSearch, groupSearchBase, groupSearchFilter, groupMembershipStrategy, managerDN, managerPasswordSecret, inhibitInferRootDN, disableMailAddressResolver, cache, environmentProperties, displayNameAttributeName, mailAddressAttributeName, IdStrategy.CASE_INSENSITIVE, IdStrategy.CASE_INSENSITIVE);
     }
 
-    // BEING TODO Jenkins 1.577+
-    /**
-     * @deprecated will be removed once we depend on Jenkins 1.577+
-     */
     @DataBoundConstructor
```
jglick: BTW time for a switch to `@DataBoundSetter`!

PR #16, opened 2017-03-27
Add some real tests

stephenc:
@reviewbybees


pom.xml
```diff
@@ -52,6 +54,180 @@
...
+      <version>${apacheds.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.directory.server</groupId>
+      <artifactId>apacheds-xdbm-partition</artifactId>
+      <version>${apacheds.version}</version>
+      <scope>test</scope>
+    </dependency>
```
oleg-nenashev: Whoa, maybe there is an aggregator dependency?

src/test/java/hudson/security/LDAPRule.java
```diff
@@ -0,0 +1,481 @@
...
+import org.junit.rules.MethodRule;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.Statement;
+
+/**
+ * Starts an embedded LDAP server.
+ */
+public class LDAPRule implements TestRule, MethodRule {
```
oleg-nenashev: Ideally it belongs to a standalone test library somewhere within Apache, but LGTM

src/test/java/hudson/security/LDAPEmbeddedTest.java
```diff
@@ -0,0 +1,80 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package hudson.security;
```
oleg-nenashev: Imho all new logic should be placed within `org.jenkinsci.plugins.ldap` package to avoid confusion

pom.xml
stephenc: there is but it's borked and pulls in slf4j and bouncycastle

rsandell:
It's not nice to steal from people ;)

PR #17, opened 2017-03-27
[JENKINS-21475] Multi server/OU support

rsandell:
[JENKINS-21475](https://issues.jenkins-ci.org/browse/JENKINS-21475)


rsandell:
@reviewbybees

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -54,19 +54,13 @@
...
 import java.net.URI;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Hashtable;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
stephenc: check your IDE, no wildcard imports please

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -465,54 +472,90 @@ public LDAPSecurityRealm(String server, String rootDN, String userSearchBase, St
...
-        this.userSearch = userSearch!=null ? userSearch : DescriptorImpl.DEFAULT_USER_SEARCH;
-        this.groupSearchBase = fixEmptyAndTrim(groupSearchBase);
-        this.groupSearchFilter = fixEmptyAndTrim(groupSearchFilter);
-        this.groupMembershipStrategy = groupMembershipStrategy == null ? new FromGroupSearchLDAPGroupMembershipStrategy("") : groupMembershipStrategy;
+        this(createLdapConfiguration(server, rootDN, userSearchBase, userSearch, groupSearchBase, groupSearchFilter, groupMembershipStrategy, managerDN, managerPasswordSecret, inhibitInferRootDN, environmentProperties, displayNameAttributeName, mailAddressAttributeName),
+                disableMailAddressResolver, cache, userIdStrategy, groupIdStrategy);
+    }
+
+    @DataBoundConstructor
+    public LDAPSecurityRealm(List<LDAPConfiguration> configurations, boolean disableMailAddressResolver, CacheConfiguration cache, IdStrategy userIdStrategy, IdStrategy groupIdStrategy) {
```
stephenc: why is the mail address resolver not part of the LDAP configuration?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -537,78 +580,87 @@ public Integer getCacheTTL() {
...
+    public static String toProviderUrl(String serverUrl, String rootDN) {
         StringBuilder buf = new StringBuilder();
         boolean first = true;
-        for (String s: serverUrl.split("\\s+")) {
+        for (String s : serverUrl.split("\\s+")) {
             if (s.trim().length() == 0) continue;
             s = getProviderUrl(s, rootDN);
             if (s != null) {
-                if (first) first = false; else buf.append(' ');
+                if (first) first = false;
```
stephenc: If multi-line then use `{` always

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -634,62 +686,21 @@ private static String getProviderUrl(String server, String rootDN) {
...
-            throw new IllegalStateException("Failed to load "+fileName, e);
-        }
-        WebApplicationContext appContext = builder.createApplicationContext();
-
-        ldapTemplate = new LdapTemplate(findBean(InitialDirContextFactory.class, appContext));
-
-        if (groupMembershipStrategy != null) {
-            groupMembershipStrategy.setAuthoritiesPopulator(findBean(LdapAuthoritiesPopulator.class, appContext));
+        LDAPAuthenticationManager manager = new LDAPAuthenticationManager();
+        DelegateLDAPUserDetailsService details = new DelegateLDAPUserDetailsService();
```
stephenc: Would it make sense to not route through the delegate if there is one and only one LDAP configuration? That way existing configurations would behave as before and not deal with the delegation

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -813,45 +848,250 @@ public String getName() {
...
+                    Authentication a = delegate.delegate.authenticate(authentication);
+                    return updateUserDetails(new DelegatedLdapAuthentication(a, delegate.server));
+                } catch (BadCredentialsException e) {
+                    lastException = e;
+                }
+            }
+            if (lastException != null) {
+                throw lastException;
+            } else {
+                throw new UserMayOrMayNotExistException("This is not intentional.", authentication);
```
stephenc: What exactly do you mean with this message?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -813,45 +848,250 @@ public String getName() {
...
+            final String server;
+
+            public ManagerEntry(AuthenticationManager delegate, String server) {
+                this.delegate = delegate;
+                this.server = server;
+            }
+        }
+    }
+
+    protected static class DelegatedLdapAuthentication implements Authentication {
```
stephenc: Any reason these need to be protected?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -813,45 +848,250 @@ public String getName() {
...
+        public Authentication getDelegate() {
+            return delegate;
+        }
+
+        public String getServer() {
+            return server;
+        }
+    }
+
+    protected static class DelegatedLdapUserDetails implements LdapUserDetails, Serializable {
```
stephenc: Could this be private?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -813,45 +848,250 @@ public String getName() {
...
+                    lastUNFE = e;
+                } catch (DataAccessException e) {
+                    LOGGER.log(Level.WARNING, "LDAP connection " + delegate.server + " seems to be broken, will _not_ try the next configuration.", e);
+                    throw e;
+                }
+            }
+            if (lastUNFE != null) {
+                throw  lastUNFE;
+            } else {
+                throw new UserMayOrMayNotExistException("This is not intentional.", username);
```
stephenc: I'm guessing this is only in the case where there are no delegates... in which case you can definitively say the user is not found and those the `UsernameNotFoundException`

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -813,45 +848,250 @@ public String getName() {
...
          * {@link BasicAttributes} in LDAP tend to be bulky (about 20K at size), so interning them
          * to keep the size under control. When a programmatic client is not smart enough to
          * reuse a session, this helps keeping the memory consumption low.
          */
         private final LRUMap attributesCache = new LRUMap(32);
 
+        @Deprecated
         LDAPUserDetailsService(WebApplicationContext appContext) {
-            this(appContext, null);
+            this(appContext, null, null);
```
stephenc: missing an extra `null`

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -465,54 +472,90 @@ public LDAPSecurityRealm(String server, String rootDN, String userSearchBase, St
...
+                disableMailAddressResolver, cache, userIdStrategy, groupIdStrategy);
+    }
+
+    @DataBoundConstructor
+    public LDAPSecurityRealm(List<LDAPConfiguration> configurations, boolean disableMailAddressResolver, CacheConfiguration cache, IdStrategy userIdStrategy, IdStrategy groupIdStrategy) {
+        if (configurations == null || configurations.isEmpty()) {
+            //Correct FormException should be handled by DescriptorImpl.newInstance
+            throw new IllegalArgumentException(jenkins.security.plugins.ldap.Messages.LDAPSecurityRealm_AtLeastOne());
+        }
+        this.configurations = configurations;
```
stephenc: If the user is using a custom Groovy bean binding file we should *refuse* to accept multiple configurations, ideally forcing the UI to allow one and only one configuration.

src/test/resources/hudson/security/docker/PlanetExpressTest/PlanetExpress/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM rroemhild/test-openldap:latest
```
rsandell: The maintainer doesn't have any labels set on DockerHub, only latest exists :'(

src/main/java/hudson/security/LDAPSecurityRealm.java
rsandell: Because it is applied globally?

stephenc: yes but why is it applied globally? would it not make more sense to be scoped per server... per server I define the attribute name, so some servers I may want to turn it off and others I want to turn it on. (Ideally we don't even offer the mail address attribute if the resolver is disabled for a server)

rsandell: Well I meant globally in Jenkins.
But one other reason to turn it off is to short cut the ldap lookup. I can't really do that for the multi ou since I need to find out what configuration the user is in before I can see if the configuration for that user allows resolving emails or not. Which means an ldap search to find out if we should not do any ldap searching :)

stephenc: Seems like it would be cheap to iterate all the servers to see if at least one has the flag enabled, which would at least consolidate configuration locality.

alternatively you should document your logic for having it be a single check-box

src/main/java/hudson/security/LDAPSecurityRealm.java
rsandell: Why? in the end there will need to be beans of `LdapUserSearch`, `LdapAuthoritiesPopulator` and `AuthenticationManager` defined or `createSecurityComponents` will fail. Those beans are then wrapped by classes we control.

stephenc: because they can do whatever crazy they want in that binding... including defining a custom groovy class that may make assumptions about the classes of the different beans... it's likely a low risk, but it is still a risk... plus nudging them to stop using the custom binding (which they probably do not need any more) is a good thing.

In any case the custom binding will likely not be using any of the settings for the different servers so you will just have an array of identical search beans

src/main/java/hudson/security/LDAPSecurityRealm.java
rsandell: Nope, it is this signature that I want to call `LDAPUserDetailsService(WebApplicationContext appContext, LDAPGroupMembershipStrategy groupMembershipStrategy, String server)`

src/main/java/hudson/security/LDAPSecurityRealm.java
rsandell: Yes, here I could do that, but in `authenticate` I believe I would leak information that can be used to probe the system if I directly say that the user does not exist.

src/main/java/hudson/security/LDAPSecurityRealm.java
rsandell: _"This should never happen"_ but with other words :)

src/main/java/hudson/security/LDAPSecurityRealm.java
stephenc: that should be a concern of core not a concert of the security realm IMHO

rsandell:
Another thing remains as well; when a server returns bad credentials if there should be a check if the user exists on that server before continuing to the next.

rsandell:
@stephenc I believe I've covered your concerns now + some of my own, please rereview.

I need to add some tests for the custom bind script blocking as well, but you should be able to review the rest while I'm doing that.

mpender:
+1 on this PR, would love to have it CR'd and merged

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -893,31 +884,49 @@ public String getName() {
     }
 
     private class LDAPAuthenticationManager implements AuthenticationManager {
-
-        private final List<ManagerEntry> delegates;
+        private final List<ManagerEntry> delegates = new ArrayList<>();;
```
stephenc: :ant: `;;` !?!

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,6 +1,7 @@
 FromUserRecordLDAPGroupMembershipStrategy.DisplayName=Parse user attribute for list of LDAP groups
 FromGroupSearchLDAPGroupMembershipStrategy.DisplayName=Search for LDAP groups containing user
 LDAPSecurityRealm.AtLeastOne=At least one LDAP server configuration must be configured.
+LDAPSecurityRealm.NotSameServer=Each server configuration must have a unique server string
```
stephenc: So with this solution I cannot have two different configurations in the same LDAP server without resorting to port trickery or a DNS name alias.

This would seem to be in conflict with the use case in the original request:

> After a company merger/consolidation there is a single authentication system but the schemas have not been merged.

In that case, there would be a single LDAP "server" but it might have two distinct trees with two distinct manager passwords.

I would think that the correct solution is to use a composite key comprising the server, root DN, user search base and user filter. It may be necessary to tweak how that gets consolidated e.g. to handle the case where the combined root DN + user search base is the same effective user search base as another.

However we would need to check with @cyrille-leclerc as to whether there is a need to specify multiple configurations for the same LDAP server. The agreed acceptance criteria do not list that as necessary, but Cyrille was not part of a min-max of the AC

stephenc:
@aheritier said:

> Our ... customer ... would like to have users from two forests ...

After checking with @jtnord 

The two forests could well require different configuration (e.g. different manager password and, if optimizing the search very likely, different search base) but they may both be accessed on the same server (i.e. the global catalog)

I thus think using the server as the uniqueness key for each configuration is thus a :bug: against the requirements driving this effort. A composite key should probably be used instead.

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
rsandell: As your example above it is easy to have distinguished server strings for the same physical server; just prefix the other one with `ldap://` for example.

rsandell:
OK, So I'm merging with known issue [JENKINS-44743](https://issues.jenkins-ci.org/browse/JENKINS-44743)
And try to fix that before GA release.

rsandell:
@reviewbybees done

PR #18, opened 2017-03-27
JENKINS-43388 Provide a "Test" button that allows verification of the LDAP settings before application to prevent lockout

stephenc:
@reviewbybees


stephenc:
ldap-validate

stephenc:
Next steps are to determine what automated tests of the config can be done and render a more detailed report

HRMPW:
@stephenc should we put the result of the test in the dialog window itself? This would serve two purposes.
1.  It is easier to see the result instead of having to find it on the main page
2. If a wants to try a different username or password they won't have to re-open the dialog

stephenc:
@HRMPW The issue then is that when you close the window you lose the error message and you may need the error message to correct the setting you have

HRMPW:
What about using an "Error" instead of a "Warning" message when it doesn't work and make it red? Could we also colorize the successful one in green?  The Success is the one that is harder to see.

stephenc:
> What about using an "Error" instead of a "Warning" message when it doesn't work and make it red? Could we also colorize the successful one in green? The Success is the one that is harder to see.

Yes that is my "need to improve the server side validation" point. This far is just the UI on the client that allows the server to

stephenc:
screen shot 2017-03-27 at 17 51 59

@HRMPW here's some of the improvements... more checks will have to come

HRMPW:
Does this list have the potential to sprawl unchecked?

Is "Only basic group membership reported" an error? What does that mean?

stephenc:
That's a warning. Not an error. If we got no groups that would be an error.

If we get more than just "authenticated" that's ok.

If we just get "authenticated" we could be dealing with a user who is not in any groups... or dealing with misconfigured group lookup... hence warning.

And there are two code paths that need checking... login and lookup!

stephenc:
ldap-validate

Latest validation checks demo

stephenc:
ATH tests: https://github.com/stephenc/acceptance-test-harness/tree/ldap-better-validation

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
...
+LDAPSecurityRealm.AvailableAttributes=Available attributes are:
+LDAPSecurityRealm.UserLookupManagerDnRequired=Is a Manager DN and password required to lookup user details?
+LDAPSecurityRealm.UserLookupManagerDnCorrect=Is a Manager DN and password correct?
+LDAPSecurityRealm.UserLookupManagerDnPermissions=Does the Manager DN have permissions to perform user lookup?
+LDAPSecurityRealm.GroupLookupManagerDnPermissions=Does the Manager DN have permissions to perform group lookup?
+LDAPSecurityRealm.GroupLookupManagerDnRequired=Is a Manager DN and password required to lookup group details?
+LDAPSecurityRealm.UserLookupSettingsCorrect=Are the user search base and user search filter settings correct?
+LDAPSecurityRealm.GroupLookupSettingsCorrect=Are the group search base and group search filter settings correct?
+LDAPSecurityRealm.NoGroupMembership=No group membership reported
+LDAPSecurityRealm.BasicGroupMembership=No LDAP group membership reported.
```
cyrille-leclerc: "reported" means "found"?
We think of  "No LDAP group membership found".

What's the difference with "LDAPSecurityRealm.NoGroupMembership"? what kind of other group is there? Do we mean that the user is not part of the "AUTHENTICATED" group?

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
...
+LDAPSecurityRealm.BasicGroupMembershipDetail=If the user is a member of some LDAP groups then the group membership \
+  settings are probably configured incorrectly.
+LDAPSecurityRealm.GroupLookupNotPossible=Group lookup: could not verify.
+LDAPSecurityRealm.GroupLookupNotPossibleDetail=Please try with a user that is a member of at least one group.
```
cyrille-leclerc: > "of at least one group" 
We mean "LDAP group"?

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
...
+LDAPSecurityRealm.UserLookupManagerDnPermissions=Does the Manager DN have permissions to perform user lookup?
+LDAPSecurityRealm.GroupLookupManagerDnPermissions=Does the Manager DN have permissions to perform group lookup?
+LDAPSecurityRealm.GroupLookupManagerDnRequired=Is a Manager DN and password required to lookup group details?
+LDAPSecurityRealm.UserLookupSettingsCorrect=Are the user search base and user search filter settings correct?
+LDAPSecurityRealm.GroupLookupSettingsCorrect=Are the group search base and group search filter settings correct?
+LDAPSecurityRealm.NoGroupMembership=No group membership reported
+LDAPSecurityRealm.BasicGroupMembership=No LDAP group membership reported.
+LDAPSecurityRealm.BasicGroupMembershipDetail=If the user is a member of some LDAP groups then the group membership \
+  settings are probably configured incorrectly.
+LDAPSecurityRealm.GroupLookupNotPossible=Group lookup: could not verify.
```
cyrille-leclerc: > "could not verify"

We mean "cloud not verify LDAP group membership"?

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
...
+LDAPSecurityRealm.GroupMembershipMatch=User groups consistent (login and lookup)
+LDAPSecurityRealm.GroupMembershipMismatch=User groups inconsistent (login versus lookup)
+LDAPSecurityRealm.GroupLookupSuccessful=Group lookup: successful ({0,choice,1#{0} group|1<{0,number,integer} groups})
+LDAPSecurityRealm.NoDisplayName=No display name specified for user!
+LDAPSecurityRealm.EmptyDisplayName=Retrieved display name was empty!
+LDAPSecurityRealm.CouldNotRetrieveDisplayName=Could not retrieve the display name attribute.
+LDAPSecurityRealm.NoEmailAddress=No email address specified for user!
+LDAPSecurityRealm.EmptyEmailAddress=Retrieved email address was empty!
+LDAPSecurityRealm.CouldNotRetrieveEmailAddress=Could not retrieve the email address attribute.
+LDAPSecurityRealm.IsAttributeNameCorrect=Is the attribute name "{0}" correct?
```
cyrille-leclerc: "attribute" -> "LDAP attribute"

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
...
+LDAPSecurityRealm.AvailableAttributes=Available attributes are:
```
cyrille-leclerc: "attributes" -> "LDAP attributes"

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
...
+LDAPSecurityRealm.UserLookupManagerDnRequired=Is a Manager DN and password required to lookup user details?
+LDAPSecurityRealm.UserLookupManagerDnCorrect=Is a Manager DN and password correct?
```
cyrille-leclerc: > "Is a"
English question: shouldn't it be "Are"

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
...
+LDAPSecurityRealm.UserLookupManagerDnPermissions=Does the Manager DN have permissions to perform user lookup?
+LDAPSecurityRealm.GroupLookupManagerDnPermissions=Does the Manager DN have permissions to perform group lookup?
+LDAPSecurityRealm.GroupLookupManagerDnRequired=Is a Manager DN and password required to lookup group details?
```
cyrille-leclerc: > "Is a"
English question: shouldn't it be "Are"

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
...
+LDAPSecurityRealm.UserLookupSettingsCorrect=Are the user search base and user search filter settings correct?
+LDAPSecurityRealm.GroupLookupSettingsCorrect=Are the group search base and group search filter settings correct?
+LDAPSecurityRealm.NoGroupMembership=No group membership reported
```
cyrille-leclerc: See question below

src/main/resources/hudson/security/LDAPSecurityRealm/config.properties
```diff
@@ -0,0 +1,6 @@
+validationBlurb=<p>Use your username and password to test your LDAP settings before saving.\
```
cyrille-leclerc: "Use your username and password" -> "Use your LDAP username and password"

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
 FromUserRecordLDAPGroupMembershipStrategy.DisplayName=Parse user attribute for list of groups
-FromGroupSearchLDAPGroupMembershipStrategy.DisplayName=Search for groups containing user
\ No newline at end of file
+FromGroupSearchLDAPGroupMembershipStrategy.DisplayName=Search for groups containing user
```
cyrille-leclerc: "groups" -> "LDAP groups"

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -1,2 +1,50 @@
...
+LDAPSecurityRealm.LoginHeader=Login
+LDAPSecurityRealm.AuthenticationSuccessful=Authentication: successful
+LDAPSecurityRealm.AuthenticationFailed=Authentication: failed for user "{0}"
+LDAPSecurityRealm.AuthenticationFailedEmptyPass=Authentication: failed for user "{0}" with empty password
+LDAPSecurityRealm.UserId=User ID: {0}
+LDAPSecurityRealm.UserDn=User Dn: {0}
+LDAPSecurityRealm.UserDisplayName=User Display Name: {0}
+LDAPSecurityRealm.UserEmail=User email: {0}
+LDAPSecurityRealm.GroupMembership=Group membership:
```
cyrille-leclerc: Isn't it only the "LDAP group membership"? The code is skipping the AUTHENTICATED_AUTHORITY group.

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
stephenc: This is no group membership at all

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
stephenc: There are two strategies for group membership. One involves searching for groups. The other is where we are just provided with a list of group names in the user record.

As such, "found" only works for 50% of the cases but "reported" works for 100%

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
stephenc: Tricky: I am referring to "Manager DN and password" as a single object. This becomes akin to asking whether it should be "Is there a pair of shoes on your feet" or "Are there a pair of shoes on your feet"... where you are trying to decide if you are focusing on the fact that there is 1 pair or 2 shoes

PR #19, opened 2017-06-08
[JENKINS-44743] Use a compound key to identify server configurations

rsandell:
See discussion in #17 

@reviewbybees esp. @stephenc

### DIFF
--- a/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
+++ b/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
@@ -528,9 +528,6 @@ public class LDAPConfiguration extends AbstractDescribableImpl<LDAPConfiguration
         }
         rootDN = rootDN.trim();
         userSearchBase = userSearchBase.trim();
-        if (userSearchBase.endsWith(rootDN)) {
-            return userSearchBase;
-        }
         return userSearchBase + "," + rootDN;
     }
 
--- a/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
+++ b/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
@@ -43,6 +43,7 @@ import org.acegisecurity.ldap.InitialDirContextFactory;
 import org.acegisecurity.ldap.LdapTemplate;
 import org.acegisecurity.ldap.search.FilterBasedLdapUserSearch;
 import org.acegisecurity.providers.ldap.LdapAuthoritiesPopulator;
+import org.apache.commons.codec.Charsets;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.digester.Digester;
 import org.apache.commons.httpclient.util.URIUtil;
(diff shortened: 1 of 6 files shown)


rsandell:
@stephenc I remember you mentioning that we need to normalize the `userSearchBase` or perhaps `userSearch`, but my notes on the details didn't stretch as far, so I uploaded this as is and hopefully you can clarify if you feel it is still needed.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -25,6 +25,7 @@
  */
 package jenkins.security.plugins.ldap;
 
+import com.trilead.ssh2.crypto.Base64;
```
stephenc: use the one from commons-codec, don't rely on trilead as the plan will be to strip that from core eventually

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -472,6 +495,55 @@ private static String addPrefix(String server) {
         else return "ldap://" + server;
     }
 
+    private String generateId() {
+        return generateId(server, rootDN, userSearchBase, userSearch);
+    }
+
+    @Restricted(NoExternalUse.class)
+    static String generateId(String serverUrl, String rootDN, String userSearchBase, String userSearch) {
+        final MessageDigest digest = DigestUtils.getMd5Digest();
```
stephenc: Is it worth all this instead of `Util.getDigestOf(normalizeServer(serverUrl)+...+...)`

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -472,6 +495,55 @@ private static String addPrefix(String server) {
...
+        digest.update(normalizeServer(serverUrl).getBytes(Charsets.UTF_8));
+        if (StringUtils.isNotBlank(rootDN)) {
+            digest.update(rootDN.getBytes(Charsets.UTF_8)); //Should have been inferred in the constructor if needed
+        } else {
+            digest.update(new byte[]{0});
+        }
+        if (StringUtils.isNotBlank(userSearchBase)) {
```
stephenc: so these are three equivalent configurations:

* root DN `dc=example,dc=com` and user search base `dc=Users`
* root DN `dc=com` and user search base `dc=Users,dc=example`
* empty root DN (with infer root DN unchecked) and user search base `dc=Users,dc=example,dc=com`

You do not generate the same ID for these three configurations, :bug:

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
rsandell: I don't like doing too many ternary operations

stephenc: StringBuilder would be simpler code and less reliance on external dependencies for the same result

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
rsandell: done

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
rsandell: done

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -472,6 +494,69 @@ private static String addPrefix(String server) {
...
+        }
+        if (isBlank(rootDN)) {
+            return userSearchBase;
+        }
+        if (isBlank(userSearchBase)) {
+            return rootDN;
+        }
+        rootDN = rootDN.trim();
+        userSearchBase = userSearchBase.trim();
+        if (userSearchBase.endsWith(rootDN)) {
```
stephenc: Why? the rootDN is alway appended when being used IIUC

rsandell: I don't understand, it's not doing that here.

stephenc: The *actual search* will always be relative to the root DN AIUI so adding it just means the effective search is `userBase + rootDn + rootDn` which is silly

stephenc: https://github.com/jenkinsci/ldap-plugin/commit/9ff3f123515e6e5f47a1e0db6b5775c687b9dcfe should explain what I mean. If I was wrong then these two additional tests would not pass... but they pass so I am wrong and your double prepending with rootDN is a :bug:

rsandell:
@stephenc OK I've dumbed-down the normalization a bit.

PR #22, opened 2017-09-11
[Fix JENKINS-45431] Moved the config snippet to the correct jelly

rsandell:
@reviewbybees


rsandell:
The test errors seems due to BC version problem when running with the a core, not an issue in this change.

rysteboe:
Failures will be fixed by bumping the bouncycastle version, which can be done once we fix the PCT failures.  Seems to be a test issue only if it follows the trend of similar issues we've seen.

rsandell:
@reviewbybees done

PR #23, opened 2017-09-13
More logging

rsandell:
Provide more logging for when configuration/connection is wrong.

@reviewbybees


rsandell:
Build fails on newer core because bouncy castle, the main build passes.

jglick:
> Build fails on newer core because bouncy castle, the main build passes.

So, what commit/PR introduced CI failures into `master`? Who merged it anyway and why? Where is the PR fixing the master build? I do not think any other PRs to this plugin should even be considered until that is straightened out.

rysteboe:
@jglick see my comment on #22.  I don't know the background of how it got broken in the first place for this plugin, but I did test that both that PR and master were fixed once the bouncycastle version was fixed.  The meeting later today should finally iron out what we're going to do about that and then the CI build can be fixed here too.

jglick:
I presume this will need to be merged with #24.

jglick:
Please `git merge master`.

rsandell:
@jglick done

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -688,6 +689,17 @@ public LDAPConfiguration getConfigurationFor(String configurationId) {
         return null;
     }
 
+    @CheckForNull
+    private static LDAPConfiguration _getConfigurationFor(String configurationId) {
+        final Jenkins jenkins = Jenkins.getInstance();
+        final SecurityRealm securityRealm = jenkins == null ? null : jenkins.getSecurityRealm();
```
dwnusbaum: Looks like FindBugs is failing the build against 2.73.1 because of the `Jenkins#getInstance` changes:

    Redundant nullcheck of jenkins, which is known to be non-null in hudson.security.LDAPSecurityRealm._getConfigurationFor(String) [hudson.security.LDAPSecurityRealm] Redundant null check at LDAPSecurityRealm.java:[line 695] RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE

Is there a way to disable FindBugs for the 2.73.1 build? Against 1.651.3 (baseline in pom.xml) the null check is necessary.

PR #24, opened 2017-09-26
Fix test failures in CI builds

dwnusbaum:
Ever since the CI builds were moved to [ci.jenkins.io](https://ci.jenkins.io/job/Plugins/job/ldap-plugin/job/master/) all builds have failed. Previously they were being built against Jenkins 1.625.3 on [jenkins.ci.cloudbees.com](https://jenkins.ci.cloudbees.com/job/plugins/job/ldap-plugin/), which worked correctly, but now they are being tested against the Jenkins default version and 2.32.3, which requires an updated version of the bouncycastle-api plugin.

Additionally, there is a simple test failure on Windows that appears to be the result of a UI change.

     java.lang.AssertionError:` Button [Advanced...] not found
         at hudson.security.LDAPSecurityRealmTest.getButtonByText(LDAPSecurityRealmTest.java:206)
         at hudson.security.LDAPSecurityRealmTest.groupMembershipAttribute(LDAPSecurityRealmTest.java:187)

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
   <properties>
     <apacheds.version>2.0.0-M23</apacheds.version>
     <apacheds.jdbm.version>2.0.0-M3</apacheds.jdbm.version>
-    <jenkins.version>1.648</jenkins.version>
+    <jenkins.version>1.651.3</jenkins.version>
   </properties>
 
   <scm>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.25</version>
+    <version>2.30</version>
   </parent>
 
   <artifactId>ldap</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,7 @@
   <properties>
     <apacheds.version>2.0.0-M23</apacheds.version>
     <apacheds.jdbm.version>2.0.0-M3</apacheds.jdbm.version>
+    <jenkins.version>1.648</jenkins.version>
   </properties>
 
   <scm>
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>bouncycastle-api</artifactId>
-      <version>1.0.3</version>
+      <version>2.16.2</version>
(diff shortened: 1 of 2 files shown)


dwnusbaum:
Need to do more testing to fix this.

dwnusbaum:
Should be good now.

pom.xml
```diff
@@ -25,6 +25,7 @@
   <properties>
     <apacheds.version>2.0.0-M23</apacheds.version>
     <apacheds.jdbm.version>2.0.0-M3</apacheds.jdbm.version>
+    <jenkins.version>1.648</jenkins.version>
```
jglick: Why this version? Generally we prefer to use an LTS as the baseline; the next would be 1.651.x. (My preference is to depend on 1.651.3, the last in the series; some others advocate for the weekly 1.651. Whatever.)

BTW `Jenkinsfile` currently requests testing against the baseline, and 2.32.3. May as well make that 2.73.1 so we can catch possible PCT issues earlier.

pom.xml
```diff
@@ -62,7 +63,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>bouncycastle-api</artifactId>
-      <version>1.0.3</version>
+      <version>2.16.2</version>
```
jglick: I hope someone else understands the implications of this.

pom.xml
jglick: (But note that only people with write permission on the repo can test the effect of changes to `Jenkinsfile`, so perhaps that needs to wait for a follow-up.)

pom.xml
rsandell: That release version has `<jenkins.version>2.16</jenkins.version>` So I doubt this will work well.

dwnusbaum: Updating to at least bouncycastle-api 1.648 is required for Jenkins > 1.648. Between bouncycastle-api 1.648.3 and 2.16, the Bouncy Castle library (bcpkix) was split from core and moved to the api plugin, so trying to use bouncycastle-api < 2.16.0 with Jenkins >= 2.16.0 means that the Bouncy Castle library is not on the classpath. 

Since we depend on bouncycastle-api in test scope, I think this means that when testing Jenkins < 2.16 both core and the api plugin are both providing a version of bcpkix.

rysteboe: I think Devin's interpretation is correct here.

jglick:
> May as well make that 2.73.1

#25

PR #26, opened 2017-10-19
[JENKINS-21784] Add support for querying group membership

dwnusbaum:
See [JENKINS-21784](https://issues.jenkins-ci.org/browse/JENKINS-21784).

Adds support for `SecurityRealm#loadGroupByGroupname(String, boolean)` and `GroupDetails#getMembers()`.

@reviewbybees

### DIFF
--- a/src/main/java/hudson/security/LDAPSecurityRealm.java
+++ b/src/main/java/hudson/security/LDAPSecurityRealm.java
@@ -44,7 +44,7 @@ import jenkins.model.Jenkins;
 import jenkins.security.plugins.ldap.FromGroupSearchLDAPGroupMembershipStrategy;
 import jenkins.security.plugins.ldap.LDAPConfiguration;
 import jenkins.security.plugins.ldap.LDAPGroupMembershipStrategy;
-import jenkins.security.plugins.ldap.LDAPSearchUtils;
+import jenkins.security.plugins.ldap.LDAPExtendedTemplate;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.acegisecurity.AcegiSecurityException;
@@ -57,7 +57,6 @@ import org.acegisecurity.GrantedAuthorityImpl;
 import org.acegisecurity.ldap.InitialDirContextFactory;
 import org.acegisecurity.ldap.LdapDataAccessException;
 import org.acegisecurity.ldap.LdapEntryMapper;
-import org.acegisecurity.ldap.LdapTemplate;
 import org.acegisecurity.ldap.LdapUserSearch;
 import org.acegisecurity.ldap.search.FilterBasedLdapUserSearch;
 import org.acegisecurity.providers.UsernamePasswordAuthenticationToken;
@@ -880,8 +879,9 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
         for (LDAPConfiguration conf : configurations) {
             String searchBase = conf.getGroupSearchBase() != null ? conf.getGroupSearchBase() : "";
             String searchFilter = conf.getGroupSearchFilter() != null ? conf.getGroupSearchFilter() : GROUP_SEARCH;
-            LdapTemplate template = conf.getLdapTemplate();
-            GroupDetailsImpl groupDetails = (GroupDetailsImpl)LDAPSearchUtils.searchForFirstEntry(template, searchBase, searchFilter, new Object[]{groupname}, new String[]{}, new GroupDetailsMapper());
+            LDAPExtendedTemplate template = conf.getLdapTemplate();
+            GroupDetailsImpl groupDetails = (GroupDetailsImpl)template.searchForFirstEntry(searchBase, searchFilter,
+                    new Object[]{groupname}, new String[]{}, new GroupDetailsMapper());
             if (groupDetails != null) {
                 if (fetchMembers) {
                     Set<String> members = conf.getGroupMembershipStrategy().getGroupMembers(groupDetails.getDn(), conf);
--- a/src/main/java/hudson/security/LDAPSecurityRealm.java
+++ b/src/main/java/hudson/security/LDAPSecurityRealm.java
@@ -870,7 +870,6 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
             cachedGroup = null;
         }
 
-        // TODO: obtain a DN instead so that we can obtain multiple attributes later
         final GroupDetailsImpl group = cachedGroup != null
                 ? cachedGroup
                 : searchForGroupName(groupname, fetchMembers);
--- a/src/main/java/hudson/security/LDAPSecurityRealm.java
+++ b/src/main/java/hudson/security/LDAPSecurityRealm.java
@@ -44,6 +44,7 @@ import jenkins.model.Jenkins;
 import jenkins.security.plugins.ldap.FromGroupSearchLDAPGroupMembershipStrategy;
 import jenkins.security.plugins.ldap.LDAPConfiguration;
 import jenkins.security.plugins.ldap.LDAPGroupMembershipStrategy;
+import jenkins.security.plugins.ldap.LDAPSearchUtils;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
 import org.acegisecurity.AcegiSecurityException;
@@ -54,7 +55,6 @@ import org.acegisecurity.BadCredentialsException;
 import org.acegisecurity.GrantedAuthority;
 import org.acegisecurity.GrantedAuthorityImpl;
 import org.acegisecurity.ldap.InitialDirContextFactory;
-import org.acegisecurity.ldap.LdapCallback;
 import org.acegisecurity.ldap.LdapDataAccessException;
 import org.acegisecurity.ldap.LdapEntryMapper;
 import org.acegisecurity.ldap.LdapTemplate;
@@ -85,15 +85,10 @@ import org.springframework.web.context.WebApplicationContext;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
-import javax.naming.InvalidNameException;
-import javax.naming.NamingEnumeration;
 import javax.naming.NamingException;
 import javax.naming.directory.Attribute;
 import javax.naming.directory.Attributes;
 import javax.naming.directory.BasicAttributes;
-import javax.naming.directory.DirContext;
-import javax.naming.directory.SearchControls;
-import javax.naming.directory.SearchResult;
 import javax.naming.ldap.Control;
 import javax.naming.ldap.LdapName;
 import java.io.IOException;
@@ -401,7 +396,7 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
     /**
      * The group details cache.
      */
-    private transient Map<String,CacheEntry<GroupDetails>> groupDetailsCache = null;
+    private transient Map<String,CacheEntry<GroupDetailsImpl>> groupDetailsCache = null;
 
     @Deprecated @Restricted(NoExternalUse.class)
     private transient Map<String,String> extraEnvVars;
@@ -845,14 +840,19 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
     @Override
     public GroupDetails loadGroupByGroupname(String groupname, boolean fetchMembers) throws UsernameNotFoundException, DataAccessException {
         groupname = fixGroupname(groupname);
-        GroupDetails cachedGroup;
+        GroupDetailsImpl cachedGroup;
         if (cache != null) {
-            final CacheEntry<GroupDetails> cached;
+            final CacheEntry<GroupDetailsImpl> cached;
             synchronized (this) {
                 cached = groupDetailsCache != null ? groupDetailsCache.get(groupname) : null;
             }
             if (cached != null && cached.isValid()) {
-                cachedGroup = cached.getValue();
+                GroupDetailsImpl cachedValue = cached.getValue();
+                if (!fetchMembers || cachedValue.getMembers() != null) {
+                    cachedGroup = cachedValue;
+                } else {
+                    cachedGroup = null;
+                }
             } else {
                 cachedGroup = null;
             }
@@ -861,8 +861,7 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
         }
 
         // TODO: obtain a DN instead so that we can obtain multiple attributes later
-
-        final GroupDetails group = cachedGroup != null
+        final GroupDetailsImpl group = cachedGroup != null
                 ? cachedGroup
                 : searchForGroupName(groupname, fetchMembers);
         if (cache != null && cachedGroup == null) {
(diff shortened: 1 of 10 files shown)


src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -830,50 +839,98 @@ public LdapUserDetails updateUserDetails(LdapUserDetails d) {
...
     @Override
     public GroupDetails loadGroupByGroupname(String groupname) throws UsernameNotFoundException, DataAccessException {
+        return loadGroupByGroupname(groupname, false);
+    }
+
+    @Override
+    public GroupDetails loadGroupByGroupname(String groupname, boolean fetchMembers) throws UsernameNotFoundException, DataAccessException {
         groupname = fixGroupname(groupname);
-        Set<String> cachedGroups;
+        GroupDetails cachedGroup;
```
dwnusbaum: It wasn't clear to me why the cache was a `Set` when the only use of the value was calling `.iterator().next()`. If I am missing something important please let me know and I can revert this part of the change.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -830,50 +839,98 @@ public LdapUserDetails updateUserDetails(LdapUserDetails d) {
...
 
-    private Set<String> searchForGroupName(String groupname) {
-        Set<String> groups = new TreeSet<>();
+    private @Nonnull GroupDetails searchForGroupName(String groupname, boolean fetchMembers) {
         for (LDAPConfiguration conf : configurations) {
             String searchBase = conf.getGroupSearchBase() != null ? conf.getGroupSearchBase() : "";
             String searchFilter = conf.getGroupSearchFilter() != null ? conf.getGroupSearchFilter() : GROUP_SEARCH;
-            groups.addAll(conf.getLdapTemplate().searchForSingleAttributeValues(searchBase, searchFilter, new String[]{groupname}, "cn"));
+            if (fetchMembers) {
+                return (GroupDetails)searchForFirstEntry(conf.getLdapTemplate(), searchBase, searchFilter, new String[]{groupname}, new LdapGroupDetailsMapper());
```
dwnusbaum: We could also use `LdapTemplate#searchForSingleEntry` so that an error is thrown if multiple groups match the search instead of picking one arbitrarily to use, but this matches the behavior of picking an arbitrary group when `fetchMembers == false`.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -892,14 +949,64 @@ public DescriptorImpl getDescriptor() {
...
+            Set<String> members = new TreeSet<>();
+            while (enumeration.hasMore()) {
+                String memberDn = String.valueOf(enumeration.next());
+                if (expectingUidInsteadOfDn) {
+                    members.add(fixUsername(memberDn));
+                } else {
+                    try {
+                        LdapName memberName = new LdapName(memberDn);
+                        Attributes memberAttributes = memberName.getRdn(memberName.size() - 1).toAttributes();
+                        if (memberAttributes.get("cn") != null) {
```
dwnusbaum: If we want to use the `Display Name LDAP attribute` option set in the LDAP plugin configuration we would need to make an extra query to get all of the attributes for the user. Is the current code an acceptable trade-off for performance, or is it worth the extra query to get the configured display name?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -830,50 +839,98 @@ public LdapUserDetails updateUserDetails(LdapUserDetails d) {
...
         groupname = fixGroupname(groupname);
-        Set<String> cachedGroups;
+        GroupDetails cachedGroup;
         if (cache != null) {
-            final CacheEntry<Set<String>> cached;
+            final CacheEntry<GroupDetails> cached;
             synchronized (this) {
                 cached = groupDetailsCache != null ? groupDetailsCache.get(groupname) : null;
             }
             if (cached != null && cached.isValid()) {
```
dwnusbaum: Should the value in the cache be invalid if it was created with `fetchMembers == false` and the current call to `loadGroupByGroupname` has `fetchMembers == true`?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -830,50 +839,98 @@ public LdapUserDetails updateUserDetails(LdapUserDetails d) {
...
+                return (GroupDetails)searchForFirstEntry(conf.getLdapTemplate(), searchBase, searchFilter, new String[]{groupname}, new LdapGroupDetailsMapper());
+            } else {
+                Set<String> groups = conf.getLdapTemplate().searchForSingleAttributeValues(searchBase, searchFilter, new String[]{groupname}, "cn");
+                if (!groups.isEmpty()) {
+                    return new GroupDetailsImpl(fixGroupname(groups.iterator().next()));
+                }
+            }
         }
-        return groups;
+        throw new UsernameNotFoundException(groupname);
```
jglick: Clearer to declare even the unchecked exceptions in `throws` of this method.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -830,50 +839,98 @@ public LdapUserDetails updateUserDetails(LdapUserDetails d) {
...
+                            dn.append(nameInNamespace);
+                        }
+
+                        return mapper.mapAttributes(dn.toString(), searchResult.getAttributes());
+                    }
+                } finally {
+                    results.close();
+                }
+
+                return null;
```
jglick: Is this going to return null out of `searchForGroupName`? (which is illegal)

src/test/java/hudson/security/LDAPEmbeddedTest.java
```diff
@@ -245,7 +245,35 @@ public void groupLookup() throws Exception {
...
+                false,
+                false,
+                new LDAPSecurityRealm.CacheConfiguration(100, 1000),
+                new LDAPSecurityRealm.EnvironmentProperty[0],
+                "cn",
+                null,
+                IdStrategy.CASE_INSENSITIVE,
+                IdStrategy.CASE_INSENSITIVE)
+        );
+        GroupDetails groupDetails = r.jenkins.getSecurityRealm().loadGroupByGroupname("HMS Bounty", true);
```
jglick: Probably useful to also test what happens when `false` is passed (or the parameter is omitted), unless an existing test is already doing this.

src/main/java/hudson/security/LDAPSecurityRealm.java
stephenc: yes the cache should be invalid in that case

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -830,50 +839,98 @@ public LdapUserDetails updateUserDetails(LdapUserDetails d) {
...
     }
 
-    private Set<String> searchForGroupName(String groupname) {
-        Set<String> groups = new TreeSet<>();
+    private @Nonnull GroupDetails searchForGroupName(String groupname, boolean fetchMembers) {
         for (LDAPConfiguration conf : configurations) {
             String searchBase = conf.getGroupSearchBase() != null ? conf.getGroupSearchBase() : "";
             String searchFilter = conf.getGroupSearchFilter() != null ? conf.getGroupSearchFilter() : GROUP_SEARCH;
-            groups.addAll(conf.getLdapTemplate().searchForSingleAttributeValues(searchBase, searchFilter, new String[]{groupname}, "cn"));
+            if (fetchMembers) {
```
stephenc: So I believe this will be invalid for the case where a different `LDAPGroupMembershipStrategy` is configured.

This change - from what I can see - seems to be assuming the default `LDAPGroupMembershipStrategy`. I think the findMembers logic needs to be implemented in `LDAPGroupMembershipStrategy` so that those where membership is by a user having an attribute referencing the group (rather than the group having an attribute referencing the user) can trigger the correct search

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -892,14 +949,64 @@ public DescriptorImpl getDescriptor() {
...
+        }
+    }
+
+    private static class LdapGroupDetailsMapper implements LdapEntryMapper {
+        @Override
+        public Object mapAttributes(String dn, Attributes attributes) throws NamingException {
+            String cn = fixGroupname(String.valueOf(attributes.get("cn").get()));
+            NamingEnumeration<?> enumeration;
+            boolean expectingUidInsteadOfDn = false;
+            if (attributes.get("member") != null) {
```
stephenc: Hardcoded assumption on one specific implementation of `LDAPGroupMembershipStrategy`

src/main/java/hudson/security/LDAPSecurityRealm.java
dwnusbaum: Yes you're right. I'll add the logic to GroupMembershipStrategy and then use the selected strategy here instead.

src/test/java/hudson/security/LDAPEmbeddedTest.java
dwnusbaum: [LDAPEmbeddedTest#groupLookup](https://github.com/dwnusbaum/ldap-plugin/blob/38176866f1f38b486d572dff54975ff89a35a769/src/test/java/hudson/security/LDAPEmbeddedTest.java#L245) omits the parameter.

src/main/java/hudson/security/LDAPSecurityRealm.java
dwnusbaum: Ok, I'll update the `throws` clause.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -830,50 +834,63 @@ public LdapUserDetails updateUserDetails(LdapUserDetails d) {
...
             } else {
-                cachedGroups = null;
+                cachedGroup = null;
             }
         } else {
-            cachedGroups = null;
+            cachedGroup = null;
         }
 
         // TODO: obtain a DN instead so that we can obtain multiple attributes later
```
dwnusbaum: `GroupDetailsImpl` now has the Group's  DN. Should I delete the TODO?

src/main/java/jenkins/security/plugins/ldap/FromGroupSearchLDAPGroupMembershipStrategy.java
```diff
@@ -76,4 +97,41 @@ public String getDisplayName() {
...
+        public Set<String> mapAttributes(String dn, Attributes attributes) throws NamingException {
+            NamingEnumeration<?> enumeration;
+            boolean expectingUidInsteadOfDn = false;
+            if (attributes.get("member") != null) {
+                enumeration = attributes.get("member").getAll();
+            } else if (attributes.get("uniqueMember") != null) {
+                enumeration = attributes.get("uniqueMember").getAll();
+            } else if (attributes.get("memberUid") != null) {
+                enumeration = attributes.get("memberUid").getAll();
+                expectingUidInsteadOfDn = true;
```
dwnusbaum: Same as above regarding making this configurable.

src/main/java/jenkins/security/plugins/ldap/FromGroupSearchLDAPGroupMembershipStrategy.java
```diff
@@ -68,6 +82,13 @@ public void setAuthoritiesPopulator(LdapAuthoritiesPopulator authoritiesPopulato
         return getAuthoritiesPopulator().getGrantedAuthorities(ldapUser);
     }
 
+    @Override
+    public Set<String> getGroupMembers(String groupDn, LDAPConfiguration conf) throws DataAccessException {
+        LdapTemplate template = conf.getLdapTemplate();
+        String[] memberAttributes = { "member", "uniqueMember", "memberUid" };
```
dwnusbaum: I think this covers the majority of member attributes on groups. Thoughts on whether the hardcoding here is ok or if this should be configurable by the user?

src/main/java/jenkins/security/plugins/ldap/FromUserRecordLDAPGroupMembershipStrategy.java
```diff
@@ -51,6 +55,7 @@
 public class FromUserRecordLDAPGroupMembershipStrategy extends LDAPGroupMembershipStrategy {
 
     private static final Logger LOGGER = Logger.getLogger(FromUserRecordLDAPGroupMembershipStrategy.class.getName());
+    private static final String USER_SEARCH_FILTER = "({0}={1})";
```
dwnusbaum: Parameters are substituted so this becomes `(getAttributeName()=groupDN)`. We could restrict it by adding and `&` with `(objectclass=inetOrgPerson)`, but I'm not sure if that would work with ActiveDirectory. Should this be configurable by the user? They have already configured `getAttributeName()`, but maybe they need specific `objectclass` filters?

src/main/java/jenkins/security/plugins/ldap/LDAPSearchUtils.java
```diff
@@ -0,0 +1,176 @@
...
+import org.acegisecurity.ldap.LdapCallback;
+import org.acegisecurity.ldap.LdapDataAccessException;
+import org.acegisecurity.ldap.LdapEntryMapper;
+import org.acegisecurity.ldap.LdapTemplate;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.springframework.dao.DataAccessException;
+
+@Restricted(NoExternalUse.class)
+public class LDAPSearchUtils {
```
dwnusbaum: This class just has static helper methods. It could instead be made similar to LdapTemplate class with instance methods. Does anyone have a strong opinion on this?

src/main/java/jenkins/security/plugins/ldap/LDAPGroupMembershipStrategy.java
```diff
@@ -64,4 +68,15 @@ public void setAuthoritiesPopulator(LdapAuthoritiesPopulator authoritiesPopulato
...
+    /**
+     * Returns a {@link Set} of all members in the specified group.
+     *
+     * @param groupDn the DN of the group whose members will be returned.
+     * @param conf the {@link LDAPConfiguration} that controls some search variables.
+     * @return a set of all members in the specified group.
+     * @throws UsernameNotFoundException if a group with the specified DN does not exist.
+     * @throws DataAccessException if there is an error performing the search.
+     */
+    public abstract @CheckForNull Set<String> getGroupMembers(String groupDn, LDAPConfiguration conf) throws DataAccessException;
```
dwnusbaum: Do I need to make this non-abstract to avoid breaking source compatibility with any third-party implementations of `LDAPGroupMembershipStrategy`?

dwnusbaum:
@stephenc I've moved the logic into the implementations of `LDAPGroupMembershipStrategy`, could you take another look?

src/main/java/jenkins/security/plugins/ldap/LDAPGroupMembershipStrategy.java
dwnusbaum: I think I should also add `@since 1.18`

dwnusbaum:
CI failure seems to be a transient network issue:

    Caused by: org.eclipse.aether.transfer.ArtifactNotFoundException: Could not find artifact org.jenkins-ci.tools:maven-hpi-plugin:jar:2.1 in azure (https://repo.azure.jenkins.io/public/)

dwnusbaum:
Looks like the CI build timed out on `windows-8-2.73` because of network issues. I'm double-checking to make sure that was the cause:

    Nov 02, 2017 2:37:11 PM org.apache.maven.wagon.providers.http.httpclient.impl.execchain.RetryExec execute
    INFO: Retrying request to {s}->https://repo.maven.apache.org:443
    Nov 02, 2017 2:52:49 PM org.apache.maven.wagon.providers.http.httpclient.impl.execchain.RetryExec execute
    INFO: I/O exception (java.net.SocketException) caught when processing request to {s}->https://repo.maven.apache.org:443: Connection timed out (Read failed)

dwnusbaum:
Will merge with jenkinsci/ldap-plugin/#27 once that is reviewed to try again for a passing build without infra errors.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -830,50 +833,64 @@ public LdapUserDetails updateUserDetails(LdapUserDetails d) {
...
             String searchBase = conf.getGroupSearchBase() != null ? conf.getGroupSearchBase() : "";
             String searchFilter = conf.getGroupSearchFilter() != null ? conf.getGroupSearchFilter() : GROUP_SEARCH;
-            groups.addAll(conf.getLdapTemplate().searchForSingleAttributeValues(searchBase, searchFilter, new String[]{groupname}, "cn"));
+            LDAPExtendedTemplate template = conf.getLdapTemplate();
+            GroupDetailsImpl groupDetails = (GroupDetailsImpl)template.searchForFirstEntry(searchBase, searchFilter,
+                    new Object[]{groupname}, new String[]{}, new GroupDetailsMapper());
+            if (groupDetails != null) {
+                if (fetchMembers) {
+                    Set<String> members = conf.getGroupMembershipStrategy().getGroupMembers(groupDetails.getDn(), conf);
+                    groupDetails = new GroupDetailsImpl(groupDetails.getDn(), groupDetails.getName(), members);
```
rsandell: Might need to store the configuration id from where this group came from for future reference. Like what is done with the user details?

src/main/java/hudson/security/LDAPSecurityRealm.java
rsandell: sure

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -830,50 +833,64 @@ public LdapUserDetails updateUserDetails(LdapUserDetails d) {
...
 
-    private Set<String> searchForGroupName(String groupname) {
-        Set<String> groups = new TreeSet<>();
+    private @Nonnull GroupDetailsImpl searchForGroupName(String groupname, boolean fetchMembers) throws UsernameNotFoundException, DataAccessException {
         for (LDAPConfiguration conf : configurations) {
             String searchBase = conf.getGroupSearchBase() != null ? conf.getGroupSearchBase() : "";
             String searchFilter = conf.getGroupSearchFilter() != null ? conf.getGroupSearchFilter() : GROUP_SEARCH;
-            groups.addAll(conf.getLdapTemplate().searchForSingleAttributeValues(searchBase, searchFilter, new String[]{groupname}, "cn"));
+            LDAPExtendedTemplate template = conf.getLdapTemplate();
+            GroupDetailsImpl groupDetails = (GroupDetailsImpl)template.searchForFirstEntry(searchBase, searchFilter,
```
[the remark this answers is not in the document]
dwnusbaum: Ok, I'll make that change.

src/main/java/jenkins/security/plugins/ldap/FromGroupSearchLDAPGroupMembershipStrategy.java
```diff
@@ -68,6 +81,13 @@ public void setAuthoritiesPopulator(LdapAuthoritiesPopulator authoritiesPopulato
         return getAuthoritiesPopulator().getGrantedAuthorities(ldapUser);
     }
 
+    @Override
+    public Set<String> getGroupMembers(String groupDn, LDAPConfiguration conf) throws DataAccessException {
+        LDAPExtendedTemplate template = conf.getLdapTemplate();
+        String[] memberAttributes = { "member", "uniqueMember", "memberUid" };
```
dwnusbaum: (Resurrecting outdated comment). I think this covers the majority of member attributes on groups. Thoughts on whether the hardcoding here is ok or if this should be configurable by the user?

dwnusbaum:
@reviewbybees done

src/main/java/jenkins/security/plugins/ldap/FromUserRecordLDAPGroupMembershipStrategy.java
fcojfernandez: @dwnusbaum `user` and ìnetOrgPerson` are valid but selective objectclass for configured users. I recommend you to read [this article](https://social.technet.microsoft.com/wiki/contents/articles/20289.inetorgperson-class-in-active-directory.aspx)

PR #27, opened 2017-11-02
Extra logging when communication with LDAP server fails

dwnusbaum:
Picks up #23 and adds a minor change to make FindBugs happy on the CI build against 2.73.

@reviewbybees


rsandell:
:bee: on the `getActiveInstance`. The rest is my code, so I can't really review that :)

rsandell:
Put why didn't you just pull my commits? instead of what I guess is a cherry pick?

dwnusbaum:
@rsandell Yeah I cherry picked them onto master thinking there were changes since your last merge, but then realized it was unnecessary. If you'd like me to just pull your commits directly that's fine, just let me know what you prefer.

rsandell:
Well I don't think you would have had to do that even if there would be merge conflicts, lust solve those in the merge commit :)
I just thought it was strange, but now it's done so no need to redo I think.

jglick:
FTR directly pulling in the original commits is simpler and generally preferable. For example, merging this PR would automatically mark the original as merged, too.

src/test/java/jenkins/security/plugins/ldap/CaptureLogRule.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.List;
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+
+import static org.hamcrest.core.IsNot.not;
+import static org.junit.Assert.assertThat;
+
+public class CaptureLogRule extends Handler implements TestRule {
```
jglick: Would prefer that any useful functionality be added to the standard `LoggerRule` and used from there.

dwnusbaum: Ok, I'll submit a PR to augment LoggerRule.

dwnusbaum:
I've force-pushed to clean up the commit history :) I also submitted jenkins-test-harness/pull/86 to add the functionality from `CaptureLogRule` to `LoggerRule`.

jglick:
> I also submitted

But to actually check that that PR is usable, you need to `mvn -DskipTests clean deploy` its (timestamped) snapshot and consume that here in the downstream PR.

dwnusbaum:
OK I will deploy a snapshot of jenkinsci/jenkins-test-harness#86 and consume it here.

rsandell:
:bee:

jglick:
@rsandell please use the GH approve gesture

pom.xml
```diff
@@ -255,6 +255,14 @@
       <version>1.10.2</version>
       <scope>test</scope>
     </dependency>
+
+    <!-- TODO: Remove once parent has a version of jenkins-test-harness with https://github.com/jenkinsci/jenkins-test-harness/pull/86. -->
+    <dependency>
```
jglick: Rather override the property `jenkins-test-harness.version`.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -920,7 +931,11 @@ private void addDelegate(final AuthenticationManager delegate, final String conf
 
         public Authentication authenticate(Authentication authentication) throws AuthenticationException {
             if (delegates.size() == 1) {
-                return updateUserDetails(delegates.get(0).delegate.authenticate(authentication));
+                try {
+                    return updateUserDetails(delegates.get(0).delegate.authenticate(authentication));
+                } catch (AuthenticationServiceException e) {
+                    LOGGER.log(Level.WARNING, "Failed communication with ldap server.", e);
```
jglick: But you are continuing anyway? That is more than a logging change.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -940,6 +955,13 @@ public Authentication authenticate(Authentication authentication) throws Authent
                     } else {
                         lastException = e;
                     }
+                } catch (AuthenticationServiceException e) {
+                    final LDAPConfiguration configuration = getConfigurationFor(delegate.configurationId);
+                    LOGGER.log(Level.WARNING,
+                            String.format("Failed communication with ldap server %s (%s)",
+                                    delegate.configurationId, configuration != null ? configuration.getServer() : "null"),
+                            e);
+                    throw e;
```
jglick: If you are rethrowing anyway, what is the point of logging now? Presumably some caller `catch`ing the exception should be logging it already, right?

src/test/java/hudson/security/LdapMultiEmbeddedTest.java
```diff
@@ -135,4 +146,153 @@ public void loginWithBrokenServerInTheMiddle() throws Exception {
...
+    }
+
+    @Test
+    public void when_second_is_wrong_and_login_on_first_no_log() throws Exception {
+        setBadPwd(sevenSeas);
+        r.createWebClient().login("fry", "fry");
+
+        assertThat(log, not(recorded(Level.WARNING,
+                containsString(FAILED_COMMUNICATION_WITH_LDAP_SERVER),
+                CoreMatchers.<Throwable>instanceOf(DataAccessException.class))));
```
jglick: No need to match the exception if we are anyway asserting that there is no such warning, right?

src/main/java/hudson/security/LDAPSecurityRealm.java
dwnusbaum: Yes, but AIUI the goal is that someone can add a custom logger for `hudson.security.LDAPSecurityRealm` and be confident that all connection errors related to LDAP authentication will be shown there with as much information as possible (i.e. displaying the actual LDAP server that had the error). We could wrap the original exception with a more detailed message, but then the errors would not appear in the custom `hudson.security.LDAPSecurityRealm` log. What do you think?

src/main/java/hudson/security/LDAPSecurityRealm.java
dwnusbaum: I've updated the code to rethrow the exception. (It seems GitHub doesn't recognize the comment as outdated because the change is on the next line)

src/main/java/hudson/security/LDAPSecurityRealm.java
jglick: No strong opinion.

dwnusbaum:
@reviewbybees done

PR #28, opened 2017-11-10
[JENKINS-47934] Upgrade the baseline for ldap

fcojfernandez:
See [JENKINS-47934](https://issues.jenkins-ci.org/browse/JENKINS-47934)

Latest bouncycastle-api-plugin requires Jenkins >= `2.16`

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,6 +27,7 @@
     <apacheds.version>2.0.0-M23</apacheds.version>
     <apacheds.jdbm.version>2.0.0-M3</apacheds.jdbm.version>
     <jenkins.version>2.16</jenkins.version>
+    <jenkins-test-harness.version>2.32</jenkins-test-harness.version>
   </properties>
 
   <scm>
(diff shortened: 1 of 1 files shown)


dwnusbaum:
The bouncy-castle dependency was updated in #24. As noted [here](https://github.com/jenkinsci/ldap-plugin/pull/24#discussion_r141344660), since the dependency is `<scope>test</scope>` I think the only issue is that when testing against Jenkins <= 2.16 there are 2 versions of BouncyCastle on the classpath. I don't think there should be any problems for users. Have you seen errors when running this plugin on an older core? (I'm just wary of bumping the baseline on such a widely used plugin.)

fcojfernandez:
@dwnusbaum It's true that the bouncy-castle dependency is a test scope dependency. However, if we test against Jenkins <= 2.16, as the plugin declares a specific version for bouncy-castle, the dependency resolution will force the version of BouncyCastle declared for bouncy-castle:2.16.2, that is a major version than the one needed for jenkins-core, so we wouldn't testing in a real environment indeed.

fcojfernandez:
As discussed offline, I will close this PR.

PR #29, opened 2018-01-05
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`


src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1324,7 +1324,7 @@ public LdapUserDetails loadUserByUsername(String username) throws UsernameNotFou
      * If the security realm is LDAP, try to pick up e-mail address from LDAP.
      */
     @Extension
-    public static final class MailAdressResolverImpl extends MailAddressResolver {
+    public static final class MailAddressResolverImpl extends MailAddressResolver {
```
oleg-nenashev: Breaks binary compatibility though I am not sure it is used externally

rsandell:
```suggestion
    public static final class MailAdressResolverImpl extends MailAddressResolver {
```

PR #31, opened 2018-01-12
[JENKINS-48917] Add option to ignore LDAP servers if they are unavailable.

dwnusbaum:
See [JENKINS-48917](https://issues.jenkins-ci.org/browse/JENKINS-48917).

This change allows communication failures with LDAP servers (i.e., DataAccessException, AuthenticationServiceException) to be ignored so that the failed action will be reattempted with the next configuration.

@reviewbybees

### DIFF
--- a/src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/config.jelly
+++ b/src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/config.jelly
@@ -47,5 +47,8 @@
                 </f:entry>
             </f:repeatableProperty>
         </f:entry>
+        <f:entry field="ignoreIfUnavailable">
+            <f:checkbox title="${%Ignore if Unavailable}"/>
+        </f:entry>
     </f:advanced>
 </j:jelly>
--- a/src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/config.jelly
+++ b/src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/config.jelly
@@ -47,8 +47,8 @@
                 </f:entry>
             </f:repeatableProperty>
         </f:entry>
-        <f:entry field="ignoreIfUnavailable">
-            <f:checkbox title="${%Ignore if Unavailable}"/>
+        <f:entry field="ignoreIfUnavailable" title="${%Ignore if Unavailable}">
+            <f:checkbox />
         </f:entry>
     </f:advanced>
 </j:jelly>
--- a/src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/help-ignoreIfUnavailable.html
+++ b/src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/help-ignoreIfUnavailable.html
@@ -1,11 +1,18 @@
 <div>
     <p>
     If checked, this option indicates that authentication and user/group membership lookups
-    will be attempted against subsequent LDAP domains even if this domain is unavailable.
+    which fail due to communication errors (e.g., LDAP server is unreachable, manager password is
+    incorrect) will be reattempted using the next configuration. Other types of failures, such as
+    a user attempting to authenticate with an incorrect password, will not be reattempted.
+    </p>
     <p>
-    If unchecked (the default), authentication and user/group memberships for users in
-    subsequent LDAP domains will fail if this domain is unavailable.
+    If unchecked (the default), authentication and user/group membership lookups which fail due to
+    communication errors will not be reattempted using the next configuration.
+    </p>
     <p>
-    The latter is useful when there is a concern that the same username may exist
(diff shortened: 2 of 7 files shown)


src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/config.jelly
```diff
@@ -47,5 +47,8 @@
                 </f:entry>
             </f:repeatableProperty>
         </f:entry>
+        <f:entry field="ignoreIfUnavailable" title="${%Ignore if Unavailable}">
+            <f:checkbox />
+        </f:entry>
     </f:advanced>
 </j:jelly>
```
Wadeck: just a line break at the end could be nice

src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/help-ignoreIfUnavailable.html
```diff
@@ -0,0 +1,11 @@
...
+    <p>
+    If checked, this option indicates that authentication and user/group membership lookups
+    will be attempted against subsequent LDAP domains even if this domain is unavailable.
+    <p>
+    If unchecked (the default), authentication and user/group memberships for users in
+    subsequent LDAP domains will fail if this domain is unavailable.
+    <p>
+    The latter is useful when there is a concern that the same username may exist
+    across domains and should not map to the same Jenkins user.
+</div>
```
Wadeck: ditto

src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/help-ignoreIfUnavailable.html
```diff
@@ -0,0 +1,11 @@
+<div>
+    <p>
```
Wadeck: quite weird to not close the `p` tag but seems to case in each help file in the project. I would use `<br />` instead personnally to be valid with the standard.

src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/help-ignoreIfUnavailable.html
```diff
@@ -0,0 +1,11 @@
...
+    If checked, this option indicates that authentication and user/group membership lookups
+    will be attempted against subsequent LDAP domains even if this domain is unavailable.
```
Wadeck: _perhaps just my english level_ 
this sentence seems to say that if the domain is available and respond correctly the subsequent domains will *still* be lookup.
From my understanding of the initial problem, the goal is to let administrators say that a domain is somewhat optional, if it fails, we try the next one. If it succeeds, we use that information.

src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/help-ignoreIfUnavailable.html
```diff
@@ -0,0 +1,11 @@
...
+    <p>
+    If unchecked (the default), authentication and user/group memberships for users in
+    subsequent LDAP domains will fail if this domain is unavailable.
```
Wadeck: > will fail

"will not even be attemped" no ?

dwnusbaum: Yes I will reword this, I should also change `memberships` to `membership lookups`. Thanks!

src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/help-ignoreIfUnavailable.html
dwnusbaum: Good point, I will try reword this to make it very clear that it only happens in the event of failure.

src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/help-ignoreIfUnavailable.html
dwnusbaum: Yeah I had `</p>` tags at first but removed them to be consistent across the plugin. I went ahead and just added them back in 5c1c278.

src/main/resources/jenkins/security/plugins/ldap/LDAPConfiguration/config.jelly
dwnusbaum: Since this isn't a new file, my preference is to just leave it as-is and handle it in a formatting-only PR later if desired. Is that ok?

Wadeck: My concern was more concerning the help text, so no problem

dwnusbaum:
@reviewbybees done

PR #32, opened 2018-07-20
Configuration-as-Code compatibility

ndeloof:
[JCasC](https://github.com/jenkinsci/configuration-as-code-plugin/pull/392) relies on ExtensionPoint to determine the API implemented by a component and compute it's symbol name (until a `@Symbol` annotation is explicitly set)

see usage [here](https://github.com/ndeloof/configuration-as-code-plugin/commit/e60cc3d10d648acff0ef188bf9fb5763ae4079e8)

### DIFF
--- a/src/main/java/jenkins/security/plugins/ldap/LDAPGroupMembershipStrategy.java
+++ b/src/main/java/jenkins/security/plugins/ldap/LDAPGroupMembershipStrategy.java
@@ -23,6 +23,7 @@
  */
 package jenkins.security.plugins.ldap;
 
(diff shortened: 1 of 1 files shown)


reinholdfuereder:
Does this also solve the following warning log (it does not seem to cause any problems though):
```
WARNING i.j.p.casc.BaseConfigurator#createAttribute: Can't handle class jenkins.security.plugins.ldap.LDAPConfiguration#extraEnvVars: type is abstract but not Describable.
```

ndeloof:
inactive, feel free to take over

PR #34, opened 2019-01-28
[JENKINS-55813] Improve analysis of the AD attributes

Wadeck:
See [JENKINS-55813](https://issues.jenkins-ci.org/browse/JENKINS-55813).

### DIFF
--- a/src/main/java/hudson/security/UserAttributesHelper.java
+++ b/src/main/java/hudson/security/UserAttributesHelper.java
@@ -43,21 +43,21 @@ import java.util.concurrent.TimeUnit;
 @Restricted(NoExternalUse.class)
 public class UserAttributesHelper {
     // https://support.microsoft.com/en-us/help/305144/how-to-use-the-useraccountcontrol-flags-to-manipulate-user-account-pro
-    private static String ATTR_USER_ACCOUNT_CONTROL = "userAccountControl";
-    private static String ATTR_ACCOUNT_EXPIRES = "accountExpires";
+    private static final String ATTR_USER_ACCOUNT_CONTROL = "userAccountControl";
+    private static final String ATTR_ACCOUNT_EXPIRES = "accountExpires";
     // for Windows Server 2003-based domain
-    private static String ATTR_USER_ACCOUNT_CONTROL_COMPUTED = "msDS-User-Account-Control-Computed";
+    private static final String ATTR_USER_ACCOUNT_CONTROL_COMPUTED = "msDS-User-Account-Control-Computed";
     // for ADAM (Active Directory Application Mode), replace the ADS_UF_DISABLED
-    private static String ATTR_USER_ACCOUNT_DISABLED = "msDS-UserAccountDisabled";
+    private static final String ATTR_USER_ACCOUNT_DISABLED = "msDS-UserAccountDisabled";
     // for ADAM, replace the ADS_UF_PASSWORD_EXPIRED
-    private static String ATTR_USER_PASSWORD_EXPIRED = "msDS-UserPasswordExpired";
+    private static final String ATTR_USER_PASSWORD_EXPIRED = "msDS-UserPasswordExpired";
 
     // https://docs.microsoft.com/en-us/windows/desktop/adschema/a-accountexpires
     // constant names follow the code in Iads.h
-    private static long ACCOUNT_NO_EXPIRATION = 0x7FFFFFFFFFFFFFFFL;
-    private static int ADS_UF_DISABLED = 0x0002;
-    private static int ADS_UF_LOCK_OUT = 0x0010;
-    private static int ADS_UF_PASSWORD_EXPIRED = 0x800000;
+    private static final long ACCOUNT_NO_EXPIRATION = 0x7FFF_FFFF_FFFF_FFFFL;
+    private static final int ADS_UF_DISABLED = 0x0002;
+    private static final int ADS_UF_LOCK_OUT = 0x0010;
+    private static final int ADS_UF_PASSWORD_EXPIRED = 0x80_0000;
 
     public static boolean checkIfUserIsEnabled(@Nonnull Attributes user) {
         try {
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/security/UserAttributesHelper.java
```diff
@@ -0,0 +1,190 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Ease all the computations required to determine the user account optional attributes for creating
+ * the UserDetails that will be used by the SecurityRealm
+ */
+@Restricted(NoExternalUse.class)
+public class UserAttributesHelper {
+    // https://support.microsoft.com/en-us/help/305144/how-to-use-the-useraccountcontrol-flags-to-manipulate-user-account-pro
+    private static String ATTR_USER_ACCOUNT_CONTROL = "userAccountControl";
```
jvz: Do you think these fields should all be `final`?

src/main/java/hudson/security/UserAttributesHelper.java
```diff
@@ -0,0 +1,190 @@
...
+    // for Windows Server 2003-based domain
+    private static String ATTR_USER_ACCOUNT_CONTROL_COMPUTED = "msDS-User-Account-Control-Computed";
+    // for ADAM (Active Directory Application Mode), replace the ADS_UF_DISABLED
+    private static String ATTR_USER_ACCOUNT_DISABLED = "msDS-UserAccountDisabled";
+    // for ADAM, replace the ADS_UF_PASSWORD_EXPIRED
+    private static String ATTR_USER_PASSWORD_EXPIRED = "msDS-UserPasswordExpired";
+
+    // https://docs.microsoft.com/en-us/windows/desktop/adschema/a-accountexpires
+    // constant names follow the code in Iads.h
+    private static long ACCOUNT_NO_EXPIRATION = 0x7FFFFFFFFFFFFFFFL;
```
jvz:
```suggestion
    private static long ACCOUNT_NO_EXPIRATION = 0x7FFF_FFFF_FFFF_FFFFL;
```

This might help readability, though I'm not exactly sure if that syntax works with hex literals.

src/main/java/hudson/security/UserAttributesHelper.java
```diff
@@ -0,0 +1,190 @@
...
+        try {
+            String userAccountControl = getStringAttribute(user, ATTR_USER_ACCOUNT_CONTROL);
+            if (userAccountControl != null) {
+                int uacAsInt = Integer.parseInt(userAccountControl);
+                if ((uacAsInt & ADS_UF_DISABLED) == ADS_UF_DISABLED) {
+                    return false;
+                }
+            }
+
+            String adamUserAccountDisabled = getStringAttribute(user, ATTR_USER_ACCOUNT_DISABLED);
```
jvz: Is this a typo? adam?

src/main/java/hudson/security/UserAttributesHelper.java
```diff
@@ -0,0 +1,190 @@
...
+
+            return true;
+        } catch (NamingException e) {
+            return true;
+        }
+    }
+
+    // documentation: https://docs.microsoft.com/en-us/windows/desktop/adschema/a-accountexpires
+    // code inspired by https://community.oracle.com/thread/1157460
+    private static long getNowIn100NsFromJan1601() {
```
jvz: How about a name like `getWin32EpochHundredNanos` or something? Key changes being remove the `100Ns` and spell it out somehow, and the `Jan1601` is a little awkward in my opinion. Your variable names within the function are more descriptive.

src/main/java/hudson/security/UserAttributesHelper.java
Wadeck: Active directory application mode, cf comment in the `ATTR_USER_ACCOUNT_DISABLED` field. It's a particular mode for AD, not sure exactly about the details, but the standard is different for that mode.

jvz: Oh, ok, thanks!

src/main/java/hudson/security/UserAttributesHelper.java
Wadeck: That works ;-)

src/main/java/hudson/security/UserAttributesHelper.java
Wadeck: After a second reading...

> standard is different

That's exactly the root cause of my non-confidence in those PRs, the lack of standard... :D

src/main/java/hudson/security/UserAttributesHelper.java
```diff
@@ -0,0 +1,190 @@
...
+            return true;
+        } catch (NamingException e) {
+            return true;
+        }
+    }
+
+    // documentation: https://docs.microsoft.com/en-us/windows/desktop/adschema/a-accountexpires
+    // code inspired by https://community.oracle.com/thread/1157460
+    private static long getWin32EpochHundredNanos() {
+        GregorianCalendar win32Epoch = new GregorianCalendar(1601, Calendar.JANUARY, 1);
```
jvz: `ZonedDateTime.of(1601, 1, 1, 0, 0, 0, 0, ZoneOffset.UTC)` could be easier to calculate with here.

jvz:
Can we get another review from one of the previous maintainers? @olamy @jglick @stephenc @batmat @rsandell @andresrc @dwnusbaum

stephenc:
🤷‍♀️ This is beyond my ken!

src/main/java/hudson/security/UserAttributesHelper.java
Wadeck: 1) I followed the code as mentioned in the javadoc
2) I do not find the proposed one to be more readable than the one in the PR

If anybody else agree on your proposal I will change ;)

jvz: My proposal was taking a random java.time class that sounded appropriate. Calendar is much more annoying to use.

jvz:
Do you think that `LDAPSecurityRealm.LDAPAuthenticationManager.authenticate()` should be updated to check UserDetails validity as well? Or should that be the responsibility of `AbstractPasswordBasedSecurityRealm`?

jvz:
Taking a fresh look at this, I believe the configured `UserDetailsChecker` bean should be used in the `authenticate()` method based on the exception signatures provided. The `UserDetailsService` bean should only load the values for those attributes.

PR #36, opened 2019-04-24
Extract groupName by `cn` attributes instead of taking the last value from the `dn` 

carlosrodlop:
**1. Context**

Currently, the ldap plugin for the **_groupName_** takes the first string value from the DN of a group object as it can be seen [here](https://github.com/jenkinsci/ldap-plugin/blob/ldap-1.20/src/main/java/hudson/security/LDAPSecurityRealm.java#L973).

Thus, using LDAP external groups for Security Realms configuration (for example, Role-based Authorization Strategy plugin), you can search an external group by its cn but the display value is the *_groupName_*. It works fine as long as the cn is the last string in the dn, if it is not included will not. For example 

* It works fine with

```
dn: cn=ccc_c3alm-support-product-owner, ou=ccc_c3alm, support,ou=teams,ou=C3ALM,cn=CCC,o=Grupo S
objectClass: top
objectClass: groupOfNames
cn: ccc_c3alm-support-product-owner
member: uid=n057083,ou=Usuarios,o=Produban,o=Grupo S
member: uid=n212487,ou=Usuarios, o=Produban,o=Grupo S
member: uid=n065273,ou=Usuarios, o=Produban,o=Grupo S
ou: product-owner
```

* It does not work fine with

```
dn: ou=product-owner, ou=ccc_c3alm-support,ou=teams,ou=C3ALM,cn=CCC,o=Grupo S
objectClass: top
objectClass: groupOfNames
cn: ccc_c3alm-support-product-owner
member: uid=n057083,ou=Usuarios,o=Produban,o=Grupo S
member: uid=n212487,ou=Usuarios, o=Produban,o=Grupo S
member: uid=n065273,ou=Usuarios, o=Produban,o=Grupo S
ou: product-owner
```

**So this current approach of this plugin is failing for those Jenkins external groups which do not include the CN attribute in their DN**   

**2. The Aim of this PR**

According to https://ldapwiki.com/wiki/Distinguished%20Names, DN group normally include the CN value within the DN value, but I have found some LDAP trees which do not follow this suggestion. For certain user making a change in their LDAP tree might not be an option as long as is working fine with other applications. 

That's why I am suggesting offering this option: obtaining the groupName value directly from [getAttributes](https://docs.oracle.com/javase/7/docs/api/javax/naming/directory/DirContext.html#getAttributes)
(See also: https://docs.oracle.com/javase/jndi/tutorial/basics/directory/getattrs.html) instead of the DN. 

**3. Why WIP**

I would like to get feedback before implementing ldap configuration GUI updated. I have thought in a checkbox or something like that


pom.xml
```diff
@@ -9,7 +9,7 @@
   </parent>
 
   <artifactId>ldap</artifactId>
-  <version>1.20</version>
+  <version>1.21-SNAPSHOT</version>
```
fcojfernandez: This change makes me think you have checked out from the tag instead of from master. Please, take care and rebase just in case.

fcojfernandez:
The PR builder is failing because a "No space left on device" error. You should open an INFRA ticket

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -25,6 +25,7 @@
  */
 package hudson.security;
 
+import com.iwombat.util.StringUtil;
```
dwnusbaum: Appears to be unused?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -413,6 +415,9 @@ group target (CN is a reasonable default)
 
     private boolean disableRolePrefixing;
 
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Diagnostic fields are left mutable so that groovy console can be used to dynamically turn/off probes.")
+    public static boolean GET_CN_FROM_ATTRIBUTES = Boolean.getBoolean(LDAPSecurityRealm.class.getName()+".getcnfromattributes");
```
dwnusbaum: I would change the casing of the property to either `GET_CN_FROM_ATTRIBUTES` or `getCnFromAttributes`.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,11 +975,32 @@ public String getName() {
...
         }
-    }
 
+        static String extractGroupName(LdapName name, Attributes attributes) throws NamingException {
+            final String CN = "cn";
+            String groupName = StringUtils.EMPTY;
+            if (!GET_CN_FROM_ATTRIBUTES) {
+                groupName = String.valueOf(name.getRdn(name.size() - 1).getValue());
+            } else {
+                for (NamingEnumeration ae = attributes.getAll(); ae.hasMore(); ) {
```
dwnusbaum: I think you can just use `Attribute attr = attributes.get(CN)` with a null check on the result and get rid of the loop.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,11 +975,32 @@ public String getName() {
...
             LdapName name = new LdapName(dn);
-            String groupName = fixGroupname(String.valueOf(name.getRdn(name.size() - 1).getValue()));
+            String groupName = fixGroupname(extractGroupName(name, attributes));
             return new GroupDetailsImpl(dn, groupName);
         }
-    }
 
+        static String extractGroupName(LdapName name, Attributes attributes) throws NamingException {
+            final String CN = "cn";
+            String groupName = StringUtils.EMPTY;
```
dwnusbaum: nit: `""` as a constant seems much more readable to me than `StringUtils.EMPTY`.

src/test/resources/hudson/security/planetexpressExtGroups_dnWithoutCn.ldif
```diff
@@ -0,0 +1,2506 @@
...
+ HUF5gACovgFqDFYKL3Bq+VgOl3FEw+JSafqK9n1G+Rv3iDoWO5hNItS4FweJXBKMRUdA0XDEVWz
+ meMwdry5lJ8IBHiXe5/xExMJ5I+ud7fW86stfBgg94XK5RkORB6IkVRaWZ4S9bo7QCs0VUmKhRD
+ lbZCKTfXDhKhlyqy6mSFVJZ1MAX3AV0kBt1Dgd6SkThs6GXGd0gInfIcaF+WHBQV2R80A5QRwlv
+ qjmxGjN4xaRdlbp3dwT6REsR6pYOgvGFtXDGi3QzfnFRXPcPlwU1DO4lGItwYUIhiYxFaddmNPj
+ obRXFNuigXB03DOeJcd4gv1OCYAfMME5WXZCOy4VD3KfTc0/wC95keX+UV/aU+DB3zeGXnYqijP
+ 5ioKdIQeCzEHjTmpwPKkWd8QsUaXVYNWveDwmBoFUUwaJaYSxLcEG0lAHe2+JfsZ40AQFqkW7OO
+ YGVIDkYBa28uAWJs+9wBptnCeksRfwMJlGvOlWF3UIaRJtjUCgOxHYxkRKA4s0aL2QaluFBQbES
+ Hl0B2Pf4rB7SCLBsFZa8AlYfIQkVIkPvouYj1LlvRdRrLHxHufcBdREEa6FSELRw41BuXR5iQmw
+ xtSuWZ07IsQdfcbAgX4f5Yck/sTC77kGk+pYE1tu0v0sQi9bGFztWvmmFN669ZzsDbvW1/JmtYQ
+ TIiO/SturEJY
```
dwnusbaum: Can you please change these names to something random/anonymous? Seems better to not use the actual names of groups from a real company's LDAP schema.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,11 +975,32 @@ public String getName() {
...
+            if (!GET_CN_FROM_ATTRIBUTES) {
+                groupName = String.valueOf(name.getRdn(name.size() - 1).getValue());
+            } else {
+                for (NamingEnumeration ae = attributes.getAll(); ae.hasMore(); ) {
+                    Attribute attr = (Attribute) ae.next();
+                    if (CN.equals(attr.getID())) {
+                        for (NamingEnumeration e = attr.getAll(); groupName.equals(StringUtils.EMPTY);) {
```
dwnusbaum: Do you also need `e.hasMore()` in the condition?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,11 +975,32 @@ public String getName() {
...
+                            groupName = e.next().toString();
```
dwnusbaum: Javadoc for `Attribute.getAll()` says that "Each element of the enumeration is a possibly null Object", so I think you should null check this.

pom.xml
carlosrodlop: solved

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,11 +975,32 @@ public String getName() {
...
+                            if (e.hasMore()) {
+                                LOGGER.log(Level.WARNING,"The group " + name.getRdns() + " has more than one cn value. The first one  (" + groupName + ") has been assigned as external group name");
```
dwnusbaum: I would hesitate to log this at WARNING continuously. If the schema has a group like this, then every time the group is loaded this message will be printed, which seems pretty spammy. Maybe only log once, or move it to `Level.FINE` so that it can be used as a diagnostic tool if things aren't working but isn't printed otherwise.

src/main/java/hudson/security/LDAPSecurityRealm.java
carlosrodlop: done

src/main/java/hudson/security/LDAPSecurityRealm.java
carlosrodlop: done

src/test/resources/hudson/security/planetexpressExtGroups_dnWithoutCn.ldif
carlosrodlop: @dwnusbaum I don't get this point :( Can you clarify it better? Thanks

src/main/java/hudson/security/LDAPSecurityRealm.java
carlosrodlop: done

src/main/java/hudson/security/LDAPSecurityRealm.java
carlosrodlop: I have moved it to `Level.FINE`

src/test/resources/hudson/security/planetexpressExtGroups_dnWithoutCn.ldif
dwnusbaum: IIRC, `C3ALM`, `ccc_c3alm-support-developer`, `ccc_c3alm-support-product-owner`, etc. were real names from an actual company's LDAP schema that came up in relation to this ticket, but maybe I am misremembering.

src/main/java/hudson/security/LDAPSecurityRealm.java
carlosrodlop: done

src/main/java/hudson/security/LDAPSecurityRealm.java
carlosrodlop: done

src/main/java/hudson/security/LDAPSecurityRealm.java
carlosrodlop: See https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fldap-plugin/detail/PR-36/6/pipeline

> [2019-09-26T07:22:30.811Z] [INFO] Nullcheck of ae at line 987 of value previously dereferenced in hudson.security.LDAPSecurityRealm$GroupDetailsMapper.extractGroupName(LdapName, Attributes) [hudson.security.LDAPSecurityRealm$GroupDetailsMapper, hudson.security.LDAPSecurityRealm$GroupDetailsMapper] At LDAPSecurityRealm.java:[line 987]Redundant null check at LDAPSecurityRealm.java:[line 987] RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE

carlosrodlop:
I have changed the approach to get the CN always from the attributes. Ignoring the flag. Why? Because after some testing I have realized that a Dn can come without CN but the attribute CN must be always included in each entry for the LDAP schema. Otherwise, The following violation is raised after running the test

> Sep 26, 2019 10:11:31 AM jenkins.security.plugins.ldap.LDAPRule initDirectoryService
> INFO: Importing schema from hudson/security/LDAPEmbeddedTest/planetexpressExtGroups_withoutCn.ldif
> org.apache.directory.api.ldap.model.exception.LdapSchemaViolationException: ERR_279 Required attributes [cn(2.5.4.3)] not found within entry ou=technical-lead-doble-cn_field,ou=ccc_aas,ou=teams,ou=C3ALM,ou=people,dc=planetexpress,dc=com

carlosrodlop:
Prior to deleting anything, I have commented on the property  `getCnFromAttributes` and the old method `extractGroupName`

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -413,6 +414,9 @@ group target (CN is a reasonable default)
 
     private boolean disableRolePrefixing;
 
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Diagnostic fields are left mutable so that groovy console can be used to dynamically turn/off probes.")
+    //public static boolean getCnFromAttributes = Boolean.getBoolean(LDAPSecurityRealm.class.getName()+".getcnfromattributes");
```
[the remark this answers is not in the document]
rsandell: and since only line 418 is commented; `@SuppressFBWarnings` is now applied to the constructor below

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,9 +974,52 @@ public String getName() {
...
+        }*/
+
+        static String extractGroupName(LdapName name, Attributes attributes) throws NamingException {
+            final String CN = "cn";
+            boolean isCN = false;
+            String groupName = String.valueOf(name.getRdn(name.size() - 1).getValue());
+            for (NamingEnumeration ae = attributes.getAll(); ae.hasMore();) {
+                Attribute attr = (Attribute) ae.next();
+                if (CN.equals(attr.getID())) {
+                    for (NamingEnumeration e = attr.getAll(); e.hasMore()&&!isCN;) {
```
rsandell:
```suggestion
                    for (NamingEnumeration e = attr.getAll(); e.hasMore() && !isCN;) {
```

rsandell: So this will pick the last string in the list?
Why not just pick the first and not have ti loop?

rsandell: Is it just to be able to log the warning perhaps?

src/test/resources/hudson/security/planetexpressExtGroups_withCn.ldif
```diff
@@ -0,0 +1,2506 @@
...
+objectClass: organizationalPerson
+objectClass: person
+objectClass: top
+cn: Bender Bending Rodríguez
+sn: Bending Rodríguez
+description: Robot
+displayName: Bender
+employeeType: Ship's Robot
+givenName: Bender
+jpegPhoto:: /9j/4AAQSkZJRgABAQEAYABgAAD/4QBgRXhpZgAASUkqAAgAAAACADEBAgAHAAAA
```
rsandell: I don't think you need to pollute the repo with more unused data.

src/main/java/hudson/security/LDAPSecurityRealm.java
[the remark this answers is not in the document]
carlosrodlop: I'm removing it because the main proposed change is assigning the groupName from the fist CN value instead of doing it from last string value of the dn....why? it is not mandatory to have `cn` included into the `dn` string but it is mandatory to have defined a `cn` attribute for any group record of the LDAP schema

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,9 +974,52 @@ public String getName() {
...
+                            groupName = e.next().toString();
+                            if (e.hasMore()) {
+                                LOGGER.log(Level.FINE,"The group " + name.getRdns() + " has more than one cn value. The first one  (" + groupName + ") has been assigned as external group name");
+                            }
+                        }
+                    }
+                }
+            }
+            return groupName;
+        }*/
```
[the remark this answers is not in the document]
carlosrodlop: done

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,9 +974,52 @@ public String getName() {
...
+
+        static String extractGroupName(LdapName name, Attributes attributes) throws NamingException {
+            final String CN = "cn";
+            boolean isCN = false;
+            String groupName = String.valueOf(name.getRdn(name.size() - 1).getValue());
+            for (NamingEnumeration ae = attributes.getAll(); ae.hasMore();) {
+                Attribute attr = (Attribute) ae.next();
+                if (CN.equals(attr.getID())) {
+                    for (NamingEnumeration e = attr.getAll(); e.hasMore()&&!isCN;) {
+                        groupName = e.next().toString();
```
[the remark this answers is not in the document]
carlosrodlop: @fcojfernandez the explanation for not applying that suggested change is [here](https://github.com/jenkinsci/ldap-plugin/pull/36#discussion_r328481115)

src/test/resources/hudson/security/planetexpressExtGroups_withCn.ldif
carlosrodlop: done

src/main/java/jenkins/security/plugins/ldap/LDAPExtendedTemplate.java
```diff
@@ -114,7 +114,7 @@ public LDAPExtendedTemplate(InitialDirContextFactory dirContextFactory) {
                 public SearchResultEnumeration doInDirContext(DirContext ctx) throws NamingException {
                     SearchControls controls = new SearchControls();
                     controls.setSearchScope(SearchControls.SUBTREE_SCOPE);
-                    controls.setReturningAttributes(attributeNames);
+                    controls.setReturningAttributes(null);
```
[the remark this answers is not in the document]
carlosrodlop: setReturningAttributes - null indicates that all attributes will be returned

carlosrodlop:
@alecharp I can make the suggested change:

> maybe a for loop to transform into a while one.

For learning purposes, can you explain the benefit of it?

carlosrodlop:
The PR is old so it's understandable we have lost the [context ](https://github.com/jenkinsci/ldap-plugin/pull/36#issue-273090002), I include myself :) Please, read in case we have forgotten why we are trying to implement this change.

jtnord:
> dn: ou=product-owner, ou=ccc_c3alm-support,ou=teams,ou=C3ALM,cn=CCC,o=Grupo S

that does not look like it would be unique (because its leaf part is just an organisational unit).
whilst a DN does not need to have a CN leaf - as it is unique it does need to have a leaf at its most left part (which could be a uuid).  I am missing the part in the referenced link that says that you can have a non unique DN or that it needs to be somethign other than fully qualified, or that the DN retreived is not actually the DN of the object.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,9 +971,31 @@ public String getName() {
...
             return new GroupDetailsImpl(dn, groupName);
         }
+
+        static String extractGroupName(LdapName name, Attributes attributes) throws NamingException {
+            final String CN = "cn";
+            boolean isCN = false;
+            String groupName = String.valueOf(name.getRdn(name.size() - 1).getValue());
+            for (NamingEnumeration ae = attributes.getAll(); ae.hasMore();) {
+                Attribute attr = (Attribute) ae.next();
+                if (CN.equals(attr.getID())) {
```
jtnord: attribute names are case insensitive.  The API does not mention that it lower cases them and this only matches the lower case.  Not sure about the interop here.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,9 +971,31 @@ public String getName() {
...
-            String groupName = fixGroupname(String.valueOf(name.getRdn(name.size() - 1).getValue()));
+            String groupName = fixGroupname(extractGroupName(name, attributes));
             return new GroupDetailsImpl(dn, groupName);
         }
+
+        static String extractGroupName(LdapName name, Attributes attributes) throws NamingException {
+            final String CN = "cn";
+            boolean isCN = false;
+            String groupName = String.valueOf(name.getRdn(name.size() - 1).getValue());
+            for (NamingEnumeration ae = attributes.getAll(); ae.hasMore();) {
```
jtnord: seems highly inefficient when `attributes.get(CN)` will get you the attribute.  (subject to case sensitivity see below - depnds on how the attributes have been configured - but SHOULD be case insensitive for LDAP.)

carlosrodlop: done

src/test/resources/hudson/security/planetexpressExtGroups_dnWithoutCn.ldif
carlosrodlop: Good point, it has been taking into account

src/main/java/hudson/security/LDAPSecurityRealm.java
carlosrodlop: @jtnord also agreed on that point. It has been implemented,

carlosrodlop:
@rsandell can you re-review this one?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -970,9 +971,29 @@ public String getName() {
...
             return new GroupDetailsImpl(dn, groupName);
         }
+
+        static String extractGroupName(LdapName name, Attributes attributes) throws NamingException {
+            final String CN = "cn";
+            boolean isCN = false;
+            String groupName = String.valueOf(name.getRdn(name.size() - 1).getValue());
+            Attribute cnAttribute = attributes.get(CN);
+            if (cnAttribute != null) {
+                for (NamingEnumeration e = cnAttribute.getAll(); e.hasMore() && !isCN;) {
```
rsandell: This still looks unnecessarily complex to me.

carlosrodlop: Please, let me know what you have in mind to simply it.

PR #43, opened 2020-04-19
[JENKINS-61854] Get the json value depending on the Jenkins version

MRamonLeon:
My previous fix wasn't enough. On java code it's also needed to get the correct element from the json based on the Jenkins version.

Screenshot from 2020-04-21 18-33-29

Now I managed to configure a ldap server with all needed and all checks are green.


MRamonLeon:
@rsandell I hope this is the right one, sorry for the extra release :cry:

MRamonLeon:
I've tried to do a test using WebClient, but the capture of the form was getting complicated, so I gave up.

MRamonLeon:
The build failure is because docker linux images are not allowed on windows builds:

```
Step 1/1 : FROM rroemhild/test-openldap:latest
latest: Pulling from rroemhild/test-openldap
image operating system "linux" cannot be used on this platform
--->%---
```

https://docs.docker.com/docker-for-windows/#switch-between-windows-and-linux-containers

PR #46, opened 2020-07-29
Remove use of BeanBuilder; construct components in Java

jglick:
Preparation for https://github.com/jenkinsci/jenkins/pull/4848. This cleanup perhaps should have been done in #17.


jglick:
Test failures are also in `master`.

src/test/java/hudson/security/docker/MultiServerTest.java
```diff
@@ -23,12 +24,19 @@
...
+import static org.junit.Assume.assumeFalse;
+import org.junit.BeforeClass;
 
 /**
  * Tests connecting to two different servers
  */
 @LDAPTestConfiguration
 public class MultiServerTest {
+
+    @BeforeClass public static void linuxOnly() {
```
jtnord, line 36: this makes me sad as it can do it - and most devs on windows can also do it, but now the tests will all be skipped.

can you check for a `JENKINS_URL` or `BUILD_URL` in the environment so this is not skipped if docker is available and we are running in CI?

jtnord: unrelated but maybe passing  `--platform  linux/amd64` would help (or at least we could specify the requirement in the `DockerContainer` such that it could test that the platform was available locally?

jglick: Please comment in #47 instead.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -572,50 +565,80 @@ static String normalizeServer(String server) { /*package scope for testing*/
...
+    public ApplicationContext createApplicationContext(LDAPSecurityRealm realm) {
+        DefaultInitialDirContextFactory initialDirContextFactory = new DefaultInitialDirContextFactory(getLDAPURL());
+        if (getManagerDN() != null) {
+            initialDirContextFactory.setManagerDn(getManagerDN());
+            initialDirContextFactory.setManagerPassword(getManagerPassword());
+        }
+        Map<String, String> vars = new HashMap<>();
+        vars.put(Context.REFERRAL, "follow");
+        vars.put("com.sun.jndi.ldap.connect.timeout", "30000"); // timeout if no connection after 30 seconds
+        vars.put("com.sun.jndi.ldap.read.timeout", "60000"); // timeout if no response after 60 seconds
```
rsandell, line 589: It is possible that those that have customized the bean builder has done so only to change these timeouts.
It's possible we need to introduce them as system properties or something.

rsandell (approved):

PR #47, opened 2020-07-29
Suppress Dockerized tests on Windows

jglick:
These started failing regularly recently.

BTW the Dockerized tests are _flaky_ on Linux too: CI shows several recent failures like

```
starting slapd on port 389 and 636...
5ea00f09 @(#) $OpenLDAP: slapd  (Aug 10 2019 19:17:00) $
	Debian OpenLDAP Maintainers <pkg-openldap-devel@lists.alioth.debian.org>
5ea00f0a ch_calloc of 1048576 elems of 704 bytes failed
slapd: ../../../../servers/slapd/ch_malloc.c:107: ch_calloc: Assertion `0' failed.
```

### DIFF
--- a/src/test/java/hudson/security/docker/MultiServerTest.java
+++ b/src/test/java/hudson/security/docker/MultiServerTest.java
@@ -1,5 +1,6 @@
 package hudson.security.docker;
 
+import hudson.Functions;
 import hudson.model.User;
 import hudson.security.LDAPEmbeddedTest;
 import hudson.security.LDAPSecurityRealm;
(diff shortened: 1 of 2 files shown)


src/test/java/hudson/security/docker/MultiServerTest.java
```diff
@@ -23,12 +24,19 @@
...
+import org.junit.BeforeClass;
 
 /**
  * Tests connecting to two different servers
  */
 @LDAPTestConfiguration
 public class MultiServerTest {
+
+    @BeforeClass public static void linuxOnly() {
+        assumeFalse("Windows CI builders now have Docker installed…but it does not support Linux images", Functions.isWindows());
```
jtnord, line 37: this makes me sad as windows can do it - and I guess most devs on windows can also do it, but now the tests will all be skipped.

could you check for a JENKINS_URL or BUILD_URL in the environment so this is not skipped if docker is available and we are running in CI?

out of scope but as a comment ->  maybe passing `--platform linux/amd64` would help (or at least we could specify the platform requirement in the `DockerContainer` such that it could test that the platform was available locally rather than failing?

jglick: > we could specify the platform requirement in the `DockerContainer` such that it could test that the platform was available locally

Not something I am about to try. I am not even sure I have Docker for Windows still running in VirtualBox (hypervisor issues or something over my head). Feel free to contribute to `docker-fixtures`, or better yet, check whether testcontainers.org solves this and offer a patch for this and other plugins. My immediate concern here is simply to get `master` green again.

> could you check for a JENKINS_URL or BUILD_URL in the environment so this is not skipped if docker is available and we are running in CI?

I guess you mean that we are _not_ running in CI. OK.

jtnord: > I guess you mean that we are not running in CI. OK.

yup - not enough negatives

PR #48, opened 2020-07-29
Compatibility with Spring Security

jglick:
Apparently suffices to offer compatibility with https://github.com/jenkinsci/jenkins/pull/4848 while still running on current Jenkins releases.

PR #49, opened 2020-07-31
Build against Spring Security

jglick:
Following up #48 but now trying to build against https://github.com/jenkinsci/jenkins/pull/4848 and use only Spring Security.

Managed to make everything compile & pass tests. @rwinch I will likely need your help here. There seem to have been quite substantial changes between Acegi Security and Spring Security in the LDAP support, and I have not managed to find a migration guide. Most (though not all) of the problems center around some change from using `LdapUserDetails`, which used to have a `getAttributes` method, to `DirContextOperations` (which has this method) plus a `String username` but which is not an `Authentication` or `UserDetails`. I am feeling my way around in the dark, knowing next to nothing about LDAP, JNDI, etc.


src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -903,10 +895,10 @@ public GroupDetails loadGroupByGroupname(String groupname, boolean fetchMembers)
                     return groupDetails;
                 }
             // Make sure we don't throw BadCredentialsException. Catch logic matches LDAPUserDetailsService#loadUserByUsername.
-            } catch (DataAccessException e) {
```
jglick, line 907: Not really sure what the Spring Security implementation is going to throw instead of this, so just guessing.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -903,10 +895,10 @@ public GroupDetails loadGroupByGroupname(String groupname, boolean fetchMembers)
...
+            } catch (AuthenticationException e) {
                 throwUnlessConfigIsIgnorable(e, conf);
             } catch (RuntimeException e) {
-                throwUnlessConfigIsIgnorable(new LdapDataAccessException("Failed to search LDAP for group: " + groupname, e), conf);
+                throwUnlessConfigIsIgnorable(new AuthenticationServiceException("Failed to search LDAP for group: " + groupname, e), conf);
```
jglick, line 901: The most appropriate-sounding type I could find in general.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1138,23 +1131,13 @@ public DelegatedLdapUserDetails(@Nonnull LdapUserDetails userDetails, @Nonnull S
             this.configurationId = configurationId;
         }
 
-        @Override
-        public Attributes getAttributes() {
-            return userDetails.getAttributes();
-        }
-
-        @Override
-        public Control[] getControls() {
```
jglick, line 1148: Not sure what this was, but anyway it is gone now.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1317,7 +1307,7 @@ public LdapUserDetails loadUserByUsername(String username) throws UsernameNotFou
                         }
                     }
                 }
-                LdapUserDetails ldapUser = ldapSearch.searchForUser(username);
+                LdapUserDetails ldapUser = ldapSearch.searchForUser(username); // TODO mismatch between DirContextOperations+String vs. LdapUserDetails
```
jglick, line 1310: Here is one of several places where I could not figure out how to get things to even compile. If I change the type of the variable to `DirContextOperations` then some errors go away (calls to `getGrantedAuthorities` and `getAttributes`) but many others appear.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1503,36 +1477,41 @@ public Set getGroupMembershipRoles(String userDn, String username) {
...
-            return convertToUpperCase;
-        }
+        // Needed by FromGroupSearchLDAPGroupMembershipStrategy:
 
-        public String getRolePrefix() {
-            return rolePrefix;
+        private GrantedAuthority defaultRole;
+
+        @Override
+        public void setDefaultRole(String defaultRole) {
```
jglick, line 1498: Awkwardly, the supertype has

```java
	/**
	 * Returns the default role Method available so that classes extending this can
	 * override
	 * @return the default role used
	 * @see #setDefaultRole(String)
	 */
	private GrantedAuthority getDefaultRole() {
		return this.defaultRole;
	}
```

which does not make sense—it is _not_ accessible.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1503,36 +1477,41 @@ public Set getGroupMembershipRoles(String userDn, String username) {
...
-            // capture the default role in the event that legacy bean binding scripts are configuring it
-            super.setDefaultRole(defaultRole);
-            this.defaultRole = new GrantedAuthorityImpl(defaultRole);
+        public String _getRolePrefix() {
+            return getRolePrefix();
         }
+
+        public  boolean _isConvertToUpperCase() {
+            return isConvertToUpperCase();
+        }
```
jglick: `protected final` in supertype.

src/main/java/jenkins/security/plugins/ldap/BindAuthenticator2.java
```diff
@@ -40,19 +41,19 @@
      */
     private boolean hadSuccessfulAuthentication;
 
-    public BindAuthenticator2(InitialDirContextFactory initialDirContextFactory) {
-        super(initialDirContextFactory);
+    public BindAuthenticator2(BaseLdapPathContextSource contextSource) {
+        super(contextSource);
```
jglick: I guess this is the replacement??

src/main/java/jenkins/security/plugins/ldap/BindAuthenticator2.java
```diff
@@ -21,13 +21,14 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.acegisecurity.providers.ldap.authenticator;
-
-import org.acegisecurity.ldap.InitialDirContextFactory;
-import org.acegisecurity.userdetails.ldap.LdapUserDetails;
+package jenkins.security.plugins.ldap;
```
jglick: `handleBindException` is now `protected` so no more need to align packages.

src/main/java/jenkins/security/plugins/ldap/FromGroupSearchLDAPGroupMembershipStrategy.java
```diff
@@ -77,15 +77,15 @@ public void setAuthoritiesPopulator(LdapAuthoritiesPopulator authoritiesPopulato
     }
 
     @Override
-    public GrantedAuthority[] getGrantedAuthorities(LdapUserDetails ldapUser) {
-        return getAuthoritiesPopulator().getGrantedAuthorities(ldapUser);
+    public Collection<? extends GrantedAuthority> getGrantedAuthorities(DirContextOperations userData, String username) {
+        return getAuthoritiesPopulator().getGrantedAuthorities(userData, username);
```
jglick, line 81: Just trying to make something work.

src/main/java/jenkins/security/plugins/ldap/FromGroupSearchLDAPGroupMembershipStrategy.java
```diff
@@ -77,15 +77,15 @@ public void setAuthoritiesPopulator(LdapAuthoritiesPopulator authoritiesPopulato
...
     }
 
     @Override
-    public Set<String> getGroupMembers(String groupDn, LDAPConfiguration conf) throws DataAccessException {
+    public Set<String> getGroupMembers(String groupDn, LDAPConfiguration conf) {
         LDAPExtendedTemplate template = conf.getLdapTemplate();
         String[] memberAttributes = { "member", "uniqueMember", "memberUid" };
-        return (Set<String>) template.retrieveEntry(groupDn, new GroupMembersMapper(), memberAttributes);
+        return template.executeReadOnly(ctx -> new GroupMembersMapper().mapAttributes(groupDn, ctx.getAttributes(LdapUtils.getRelativeName(groupDn, ctx), memberAttributes)));
```
jglick, line 88: Just guessing here. Read only? Read write? `retrieveEntry` is nowhere to be found in Spring Security so I tried to inline the Acegi Security behavior to the extent I understood it. Note that this means `GroupMembersMapper.mapAttributes` could just be inlined here too, but I am trying to keep the diff small where possible.

src/main/java/jenkins/security/plugins/ldap/FromUserRecordLDAPGroupMembershipStrategy.java
```diff
@@ -102,37 +101,37 @@ public String getAttributeName() {
...
                     }
-                    GrantedAuthorityImpl extraAuthority = new GrantedAuthorityImpl(
-                            authoritiesPopulatorImpl.getRolePrefix() + role);
+                    GrantedAuthority extraAuthority = new SimpleGrantedAuthority(
+                            authoritiesPopulatorImpl._getRolePrefix() + role);
                     result.add(extraAuthority);
                 }
             }
-            result.addAll(authoritiesPopulatorImpl.getAdditionalRoles(ldapUser));
+            result.addAll(authoritiesPopulatorImpl.getAdditionalRoles(userData, /* TODO currently ignored anyway, but where does username come from? */null));
```
jglick, line 112: Maybe needs to be passed in as an additional argument or something?

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -578,56 +576,48 @@ static String normalizeServer(String server) { /*package scope for testing*/
 
     @Restricted(NoExternalUse.class)
     public ApplicationContext createApplicationContext(LDAPSecurityRealm realm) {
-        DefaultInitialDirContextFactory initialDirContextFactory = new DefaultInitialDirContextFactory(getLDAPURL());
+        DefaultSpringSecurityContextSource contextSource = new DefaultSpringSecurityContextSource(getLDAPURL());
```
jglick: Just guessing.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -578,56 +576,48 @@ static String normalizeServer(String server) { /*package scope for testing*/
...
         if (getManagerDN() != null) {
-            initialDirContextFactory.setManagerDn(getManagerDN());
-            initialDirContextFactory.setManagerPassword(getManagerPassword());
+            contextSource.setUserDn(getManagerDN());
+            contextSource.setPassword(getManagerPassword());
```
jglick: Inferred from `LdapAuthenticationProviderConfigurer` that this is the equivalent.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -578,56 +576,48 @@ static String normalizeServer(String server) { /*package scope for testing*/
...
         }
+        /* TODO no idea what to do with these
         Map<String, String> vars = new HashMap<>();
         vars.put(Context.REFERRAL, "follow");
         vars.put("com.sun.jndi.ldap.connect.timeout", "30000"); // timeout if no connection after 30 seconds
         vars.put("com.sun.jndi.ldap.read.timeout", "60000"); // timeout if no response after 60 seconds
         vars.putAll(getExtraEnvVars());
-        initialDirContextFactory.setExtraEnvVars(vars);
+        */
```
jglick: Could not find any equivalent to this method in `DefaultInitialDirContextFactory`.

Anyway I think `extraEnvVars` is deprecated in the Jenkins plugin. Not sure about the JNDI properties.

src/main/java/jenkins/security/plugins/ldap/LDAPExtendedTemplate.java
```diff
@@ -91,37 +85,28 @@ public LDAPExtendedTemplate(InitialDirContextFactory dirContextFactory) {
...
-                    NamingEnumeration<SearchResult> searchResults = ctx.search(base, filter, params, controls);
-                    return new SearchResultEnumeration(searchResults, mapper, getDnSuffix(base, ctx.getNameInNamespace()));
-                }
-            });
-        } catch (AuthenticationException e) {
-            throw new LdapDataAccessException("Failed to search LDAP", e);
-        }
+    private @NonNull <T> SearchResultEnumeration<T> searchForAllEntriesEnum(@NonNull final String base, @NonNull final String filter,
+            final Object[] params, final String[] attributeNames, @NonNull final LdapEntryMapper<T> mapper) {
+        return executeReadOnly(ctx -> {
```
jglick, line 103: Again, read only? Read write?

src/main/java/jenkins/security/plugins/ldap/LdapEntryMapper.java
```diff
@@ -0,0 +1,39 @@
...
+
+package jenkins.security.plugins.ldap;
+
+import javax.naming.NamingException;
+import javax.naming.directory.Attributes;
+import org.springframework.ldap.core.AttributesMapper;
+import org.springframework.ldap.core.support.AbstractContextMapper;
+
+/**
+ * Sort of like {@link AttributesMapper} but with a DN; also sort of like {@link AbstractContextMapper}.
```
jglick, line 33: Could find no exact equivalent of this Acegi Security type, but anyway it is only being used internally here.

src/main/java/hudson/security/LDAPSecurityRealm.java
Vlatombe: I've been looking into this. `LdapUserDetailsManager` seems to be a replacement for the whole method.

However, we would need a custom `UserDetailsContextMapper` in order to map `LDAPConfiguration#displayNameAttributeName` and `LDAPConfiguration#mailAddressAttributeName`

src/main/java/jenkins/security/plugins/ldap/FromGroupSearchLDAPGroupMembershipStrategy.java
Vlatombe: All the calls should be read-only. The ldap plugin never writes in the remote ldap afair.

src/main/java/hudson/security/LDAPSecurityRealm.java
jglick: Great, I will look into that.

jglick: Looked at it, but that seemed to be a way bigger refactoring, and I did not see how to fit it into the plugin’s current logic. For now, went with a simpler patch that at least fixes the types here.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1336,6 +1336,7 @@ public DelegatedLdapUserDetails loadUserByUsername(String username) throws Usern
                 // Add those, as done in LdapAuthenticationProvider.createUserDetails().
                     LdapUserDetailsImpl.Essence user = new LdapUserDetailsImpl.Essence(ldapUser);
                     user.setUsername(username);
+                    user.setDn(ldapUser.getNameInNamespace()); // otherwise the DN is missing the DC
```
jglick, line 1339: Is this right?

rsandell:
Since this will be a major version change with long sought after upgrades, I think removing some of the small obscure features is OK.

jglick:
> I think removing some of the small obscure features is OK.

Probably so, but I do not know enough about this plugin’s usage to even know which features are “small” or “obscure”.

Currently there are still quite a few test failures that are beyond the limits of my knowledge. For example: `ou=umich,ou.edu` is apparently an “invalid name” according to JNDI. This is just the search string (or “URL” or something?) configured in the test, but apparently it is supposed to be processed somehow?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1295,19 +1296,20 @@ public UserDetails loadUserByUsername(String username) throws UsernameNotFoundEx
         }
 
         @SuppressFBWarnings(value = "RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE", justification = "Only on newer core versions") //TODO remove when core is bumped
```
jglick, line 1298: Could probably clean some of this stuff up in `master` at this point, in a separate PR.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1317,63 +1319,61 @@ public LdapUserDetails loadUserByUsername(String username) throws UsernameNotFou
...
-                LdapUserDetails ldapUser = ldapSearch.searchForUser(username);
+                DirContextOperations ldapUser = ldapSearch.searchForUser(username);
                 // LdapUserSearch does not populate granted authorities (group search).
                 // Add those, as done in LdapAuthenticationProvider.createUserDetails().
-                if (ldapUser != null) {
                     LdapUserDetailsImpl.Essence user = new LdapUserDetailsImpl.Essence(ldapUser);
+                    user.setUsername(username);
+                    user.setDn(ldapUser.getNameInNamespace()); // otherwise the DN is missing the DC
 
+                    /* TODO DirContextAdapter has no setAttributes method, and anyway it looks to be of type NameAwareAttributes not BasicAttributes, so attributesCache may need to be reworked:
```
jglick, line 1329: If this cache thing is significant, need to revisit.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1409,10 +1409,10 @@ public String findMailAddressFor(User u) {
                 } else {
                     attr = DescriptorImpl.DEFAULT_MAILADDRESS_ATTRIBUTE_NAME;
                 }
-                Attribute mail = details.getAttributes().get(attr);
+                Attribute mail = DelegatedLdapUserDetails.getAttributes(details, /* TODO maybe look it up using configuration & realm */null).get(attr);
```
jglick, line 1412: Not covered by tests I guess?

src/test/java/hudson/security/LDAPEmbeddedTest.java
```diff
@@ -549,6 +548,7 @@ public void validate() throws Exception {
         assertThat("Always report outer kind as OK", validation.kind, is(FormValidation.Kind.OK));
     }
 
+    @Ignore("TODO InitialLdapContext in fact receives typesOnly, yet everything seems to work anyway in Spring (attributes are defined)")
```
jglick, line 551: Spent most of the day trying to figure this out, without success. Not clear whether there is actually something wrong in the plugin, or just the test making unwarranted assumptions.

src/test/java/hudson/security/LDAPSecurityRealmTest.java
```diff
@@ -62,6 +54,7 @@
     @Rule
     public JenkinsRule r = new JenkinsRule();
 
+    /* TODO rewrite or delete whatever this is
```
jglick, line 57: Does not even compile IIRC, and did not seem critical.

jglick:
Hmm, while functional tests pass, most ATH tests are failing with

```
java.lang.ClassCastException: com.sun.jndi.ldap.LdapCtx cannot be cast to org.springframework.ldap.core.DirContextAdapter
	at org.springframework.security.ldap.SpringSecurityLdapTemplate.searchForSingleEntryInternal(SpringSecurityLdapTemplate.java:337)
	at org.springframework.security.ldap.SpringSecurityLdapTemplate.lambda$searchForSingleEntry$2(SpringSecurityLdapTemplate.java:312)
	at org.springframework.ldap.core.LdapTemplate.executeWithContext(LdapTemplate.java:817)
	at org.springframework.ldap.core.LdapTemplate.executeReadOnly(LdapTemplate.java:803)
	at org.springframework.security.ldap.SpringSecurityLdapTemplate.searchForSingleEntry(SpringSecurityLdapTemplate.java:312)
	at org.springframework.security.ldap.search.FilterBasedLdapUserSearch.searchForUser(FilterBasedLdapUserSearch.java:128)
	at org.springframework.security.ldap.authentication.BindAuthenticator.authenticate(BindAuthenticator.java:95)
	at jenkins.security.plugins.ldap.BindAuthenticator2.authenticate(BindAuthenticator2.java:50)
	at org.springframework.security.ldap.authentication.LdapAuthenticationProvider.doAuthentication(LdapAuthenticationProvider.java:187)
	at org.springframework.security.ldap.authentication.AbstractLdapAuthenticationProvider.authenticate(AbstractLdapAuthenticationProvider.java:85)
	at org.springframework.security.authentication.ProviderManager.authenticate(ProviderManager.java:199)
	at hudson.security.LDAPSecurityRealm$LDAPAuthenticationManager.authenticate(LDAPSecurityRealm.java:1008)
	at org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter.attemptAuthentication(UsernamePasswordAuthenticationFilter.java:95)
	at org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter.doFilter(AbstractAuthenticationProcessingFilter.java:212)
	at …
```

The failing cast (from Object) is uncommented:

```java
DirContextAdapter dca = (DirContextAdapter) searchResult.getObject();
```

Seems there are two only partially related inheritance hierarchies under `DirContext`. All cryptic to me.

jglick:
Reproducible also in `hpi:run`. In `DirectoryManager`,

```java
        FactoryEnumeration factories = ResourceManager.getFactories(
            Context.OBJECT_FACTORIES, environment, nameCtx);
```

produces `null` despite the `environment` specifying `java.naming.factory.object` as `org.springframework.ldap.core.support.DefaultDirObjectFactory`. Smells like a class loading issue.

jglick:
I think I know what to fix here; just having trouble running ATH locally to confirm it.

src/main/java/hudson/security/LDAPSecurityRealm.java
rsandell: No clue, but if it works it works right? :/

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1516,22 +1521,22 @@ public boolean isGeneratingPrefixRoles() {
             return StringUtils.isNotBlank(rolePrefix) || convertToUpperCase;
         }
 
-        public boolean isConvertToUpperCase() {
+        public boolean _isConvertToUpperCase() {
```
rsandell, line 1524: Why?

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1765,7 +1770,7 @@ public FormValidation validate(LDAPSecurityRealm realm, String user, String pass
                                 .LDAPSecurityRealm_UserLookupManagerDnPermissions(),
                         jenkins.security.plugins.ldap.Messages.LDAPSecurityRealm_UserLookupSettingsCorrect());
                 // we do not flag these errors as could be probing user accounts
-            } catch (LdapDataAccessException e) {
+            } catch (AuthenticationException e) {
```
rsandell, line 1773: The old `LdapDataAccessException` meant that there was problems communicating with the ldap server, or your query was wrong or similar. Is the new `AuthenticationException` equivalent? By the name it sounds to me like it's a different exception.

src/main/java/jenkins/security/plugins/ldap/BindAuthenticator2.java
rsandell: seems like it...

src/main/java/hudson/security/LDAPSecurityRealm.java
jglick: Oh this was a bit of a mess indeed. See https://github.com/jenkinsci/ldap-plugin/pull/49#discussion_r463740875 amended by https://github.com/jenkinsci/ldap-plugin/pull/49/commits/a9dc4e48687f07fdf4ca374d55cb4ac662f6d425 (which also reduces diff size). Basically the plugin never used the `super` implementation of `convertToUpperCase` and `rolePrefix`; instead https://github.com/jenkinsci/ldap-plugin/blob/9577fe6b8ea685d04fcfe243e750eeea059a88a2/src/main/java/hudson/security/LDAPSecurityRealm.java#L1506-L1515 reimplements something like it, for reasons I do not entirely follow. So to disable the `super` implementations, but still capture values passed to the setters, while accommodating a supertype which newly made the getters (but not setters) `final`, we have these tricks.

src/main/java/hudson/security/LDAPSecurityRealm.java
jglick: There is no more `org.springframework.dao.DataAccessException` or its subtypes, so in the core PR as well as here I have converted these to the more generic `AuthenticationException`. Whether this is the most appropriate type or not, I do not know, and I am not sure we really care.

src/main/java/jenkins/security/plugins/ldap/FromUserRecordLDAPGroupMembershipStrategy.java
rsandell: isn't it the second argument to the method you are in?

src/test/java/hudson/security/LDAPEmbeddedTest.java
jglick: also https://github.com/jenkinsci/acceptance-test-harness/pull/607

rsandell: I'm actually not sure what this test is supposed to test, except judging by the name of the test it checks that environment properties are set correctly. So I don't think it specifically needs to check that you can't login after setting `java.naming.ldap.typesOnly` :) It could test some other value instead.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -742,27 +740,22 @@ private static String getProviderUrl(String server, String rootDN) {
...
     }
 
-    /**
-     * Creates security components.
-     * @return Created {@link SecurityComponents}
-     * @throws IllegalStateException Execution error
-     */
     @Override @Nonnull
     public SecurityComponents createSecurityComponents() {
         if (configurations.size() > 1) {
```
rsandell, line 745: History lesson for context;
When we implemented the multiple ldap server support we were a bit paranoid of changing the behavior in any way for configurations with only one server configured. That's why you see all of these `if (configurations.size() > 1)` all over the place. It was also the reason for the `DelegatedLdapUserDetails`  class that I see you have changed to always be used instead of before; only when multiple configurations are in effect.

Since the behavior _(both user wise but also in binary compatibility)_ is changing now and `DelegatedLdapUserDetails` is used all the time, I believe it's now possible to remove all this semi redundant code that has separate treatments for the single server configuration scenario.

But I won't block this PR if you don't want to include that cleanup work here ;).

src/test/java/hudson/security/LdapMultiEmbeddedTest.java
```diff
@@ -402,60 +400,55 @@ public void when_first_and_second_are_unavailable_and_ignorable_then_lookup_fail
                 allOf(containsString(FAILED_COMMUNICATION_WITH_LDAP_SERVER),
                         containsString(WILL_TRY_NEXT_CONFIGURATION),
                         containsString(INVALID_URL_PREFIX + planetExpress.getPort())),
-                CoreMatchers.<Throwable>instanceOf(DataAccessException.class)));
-        assertThat(log, recorded(Level.WARNING,
```
rsandell, line 406: I think the original intent here was to check that there actually was two records in the log.

jglick: Perhaps that was the intent, but in fact it was just repeating an identical assertion.

jglick: See original https://github.com/jenkinsci/ldap-plugin/blob/ff2c95a970df0c0446fe31d3b218c8bded827c2e/src/test/java/hudson/security/LdapMultiEmbeddedTest.java#L401-L410 which AFAICT was just a mistake.

jglick: See #31 by @dwnusbaum. `when_first_and_second_are_unavailable_and_ignorable_then_no_login` sensibly verifies that we have messages from both `planetExpress` and `sevenSeas`. Yet `when_first_and_second_are_unavailable_and_ignorable_then_lookup_fails`, perhaps a modified copy of that test, redundantly asserts that we have a message from `planetExpress` only.

dwnusbaum: I think it was a typo, the second assertion should have used `sevenSeas.getPort` to make sure that the plugin tried to check both servers.

src/main/java/hudson/security/LDAPSecurityRealm.java
jglick: Trying this in #52.

src/main/java/hudson/security/LDAPSecurityRealm.java
jglick: ⇒ #52

src/main/java/hudson/security/LDAPSecurityRealm.java
jglick: see also https://github.com/jenkinsci/ldap-plugin/pull/49#discussion_r495062096

src/main/java/jenkins/security/plugins/ldap/FromGroupSearchLDAPGroupMembershipStrategy.java
jglick: > could just be inlined here too

⇒ #52

rsandell (approved):

rsandell (approved):

pom.xml
```diff
@@ -28,7 +28,7 @@
     <changelist>-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
     <apacheds.version>2.0.0.AM25</apacheds.version>
-    <jenkins.version>2.235.3</jenkins.version>
+    <jenkins.version>2.257-rc30526.4bb45a702b8d</jenkins.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/4848 -->
```
jvz, line 31: To be released in the next weekly. Will this be ready to review and merge afterward?

jglick: It has already been reviewed (more reviews welcome!) and yes it would be ready to merge when the dep is on 2.266.

PR #50, opened 2020-09-10
[JENKINS-55813] Add support for LDAP password policy attributes (account locking)

jvz:
https://issues.jenkins-ci.org/browse/JENKINS-55813

This adds support for check account statuses using three common LDAP extensions for checking account validity:
* Draft LDAP password policy extension
* Active Directory account policies
* EDirectory account policies

As such, this introduces support for checking the following account states:
* Account expired or not active yet (start and end date support)
* Account lockout (used for various automated lockout mechanisms from too many auth failures)
* Account disabled (used instead of deleting an account sometimes)
* Password expired

This PR works best combined with https://github.com/jenkinsci/jenkins/pull/4925 as there would otherwise be some missing checks for these attributes.

Some of the listed scenarios have automated tests introduced based on my successes in finding ways to automate them. In particular, the password policy LDAP overlay and the Active Directory attributes are both tested.

@jeffret-b @Wadeck @daniel-beck


src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1360,6 +1360,20 @@ public LdapUserDetails loadUserByUsername(String username) throws UsernameNotFou
...
+                    }
+                    if (!UserAttributesHelper.checkIfAccountNonLocked(v)) {
+                        user.setAccountNonLocked(false);
+                    }
+                    if (!UserAttributesHelper.checkIfCredentialsAreNonExpired(v)) {
+                        user.setCredentialsNonExpired(false);
+                    }
+                    if (!UserAttributesHelper.checkIfUserIsEnabled(v)) {
+                        user.setEnabled(false);
+                    }
```
jeffret-b, line 1375: Could these be simplified to something like
user.setEnabled(UserAttributesHelper.isUserEnabled())
?

jvz: Yes, good point!

rsandell:
It doesn't look like it would be too hard to add some entries into one of the embedded tests or the docker fixture test's data and verify.

jvz:
> It doesn't look like it would be too hard to add some entries into one of the embedded tests or the docker fixture test's data and verify.

So I'm trying to manually add four more users to `planetexpress.ldif` to test out the four scenarios (using the password policy attributes at least), but I'm getting exceptions about the various attribute types not existing. Setting up this attributes was part of the automatic setup process in Apache Directory Studio when I made a test LDAP server, so I'll have to look more into this.

jvz:
Based on further testing, it seems as though this PR might have the wrong approach. These attributes are supposed to be validated LDAP server side, not client side.

jvz:
I've added an Active Directory version of the tests as well by copying some test infra from https://github.com/jenkinsci/active-directory-plugin with a small modification to the `custom.sh` script to disable a user for said test.

README.md
```diff
@@ -25,7 +25,56 @@ with API token or Jenkins CLI access but can also appear with features
...
-membership outside of a user's web session)
+membership outside of a user's web session).
+
+### Compatibility Notes
+
+Various LDAP servers use different operational attributes to make decisions on
+and expose configurations of concepts such as disabling an account, locking an
+account, and specifying a time interval the account is valid for. These policies
+are normally enforced by the LDAP server itself when performing user authentication.
+Once a user has configured API tokens or SSH keys, these parallel authentication
```
jeffret-b, line 36: This sentence needs better transition or explanation. The previous sentences are talking about LDAP servers, but this suddenly jumps, without mentioning it, to talking about Jenkins API tokens and SSH keys. The following sentence then goes back to talking about LDAP server implementation / configuration.

Maybe it needs to lead off with some sort of problem statement or explanation.

"Compatibility Notes" kind of makes sense, but only within a better context of what this whole section is about.

README.md
```diff
@@ -25,7 +25,56 @@ with API token or Jenkins CLI access but can also appear with features
...
+mechanisms can only make limited policy decisions based on checking for known user
+attributes without requiring the user to re-enter their password. These systems include
+[**slapo-ppolicy**(5)](https://linux.die.net/man/5/slapo-ppolicy) from OpenLDAP,
+Active Directory Application Mode (ADAM), classic Active Directory, and eDirectory LDAP.
+The support of these features is entirely dependent upon the LDAP server implementation
+properly exposing these operational attributes which is dependent on the exact version
+and distribution of the LDAP server in use.
```
jeffret-b, line 43: I suggest expanding this a bit to say more about how these capabilities may not work depending on whether the LDAP server implements any of them, how it implements them, how they are configured, and how they are exposed.

README.md
```diff
@@ -25,7 +25,56 @@ with API token or Jenkins CLI access but can also appear with features
...
+
+Accounts that are disabled by administrators typically expose an operational attribute
+to indicate such. The following attributes are all supported:
+
+* `pwdAccountLockedTime` value of `000001010000Z`: common LDAP attribute using password policy overlay;
+* `msDS-UserAccountDisabled` value of `TRUE`: modern [Active Directory attribute](https://docs.microsoft.com/en-us/windows/win32/adschema/a-msds-useraccountdisabled);
+* `userAccountControl` (UAC) or `msDS-User-Account-Control-Computed` with bit flag of `ADS_UF_ACCOUNTDISABLE` (0x2) present; and
+* `loginDisabled` value of `TRUE`: eDirectory attribute.
+
+#### Expired Accounts
```
jeffret-b, line 55: I really question using any time based calculations here. There are so many things that can go wrong when evaluating or comparing time. There is the question of interpreting the timestamps correctly. More significantly, there is a high likelihood of time zone differences or clock skew. If the LDAP server exists in a different time zone than the Jenkins controller, it may apply restrictions during the wrong times.

This section is rife with possibilities for being wrong and confusing things more.

src/main/java/hudson/security/UserAttributesHelper.java
```diff
@@ -0,0 +1,285 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2019, CloudBees, Inc.
```
jeffret-b, line 4: nit: Ought to go ahead and update this.

src/main/java/hudson/security/UserAttributesHelper.java
```diff
@@ -0,0 +1,285 @@
...
+        } else if (computedUac == null) {
+            return Integer.parseInt(uac);
+        } else {
+            return Integer.parseInt(uac) | Integer.parseInt(computedUac);
+        }
+    }
+
+    private static @CheckForNull Boolean getBooleanAttribute(@Nonnull Attributes user, @Nonnull String attrID) {
+        String attr = getStringAttribute(user, attrID);
+        return attr == null ? null : "true".equalsIgnoreCase(attr);
```
jeffret-b, line 260: Any reason not to use Boolean.valueOf(String) ?

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -590,8 +590,7 @@ public ApplicationContext createApplicationContext(LDAPSecurityRealm realm) {
         vars.putAll(getExtraEnvVars());
         initialDirContextFactory.setExtraEnvVars(vars);
 
-        FilterBasedLdapUserSearch ldapUserSearch = new FilterBasedLdapUserSearch(getUserSearchBase(), getUserSearch(), initialDirContextFactory);
```
jeffret-b, line 593: It doesn't look like FilterBasedLdapUserSearch is used anymore, though it is still imported and referenced several places in JavaDoc in this file.

src/main/java/jenkins/security/plugins/ldap/OperationalAttributesLdapUserSearch.java
```diff
@@ -0,0 +1,46 @@
+package jenkins.security.plugins.ldap;
```
jeffret-b, line 1: nit: Add copyright here?

jeffret-b (commented):
Looks good. The tests are nice and impressive.

I left a couple of nit comments.

The documentation needs a bit of improvement to improve user flow.

I question the reliability of the time-based checks. It would be better to leave those off. Too many things can go wrong in that area. In any event, they need better explanation, whether removed or retained.

This change probably needs an escape hatch -- some way for a Jenkins administrator to disable it. Introducing it at this point has the potential for causing problems. Maybe even these checks should be opt-in. We don't want to disable users and admins from unexpectedly being able to login. This is likely particularly important for "service" accounts, those that might be using API tokens or SSH keys, which may break because they have unexpected ldap configurations.

README.md
jvz: Timestamps have timezones specified. It's how LDAP GeneralizedTime values are stored. Are you suggesting we update the docs here or not support it at all?

src/main/java/hudson/security/UserAttributesHelper.java
jvz: Doesn't support `null`.

README.md
jvz: Added another sentence below clarifying the timezones.

jvz:
> This change probably needs an escape hatch -- some way for a Jenkins administrator to disable it. Introducing it at this point has the potential for causing problems. Maybe even these checks should be opt-in. We don't want to disable users and admins from unexpectedly being able to login. This is likely particularly important for "service" accounts, those that might be using API tokens or SSH keys, which may break because they have unexpected ldap configurations.

Do you think we should have one escape hatch for this entire feature, or individual ones for the four attributes, or some other combination?

README.md
jeffret-b: Good point. I'd still be concerned about clock skew, but that's probably not too big of a deal if we document it. I'm cautious about doing time-based comparisons based on data from another server but maybe it won't be too bad.

jeffret-b:
> Do you think we should have one escape hatch for this entire feature, or individual ones for the four attributes, or some other combination?

I don't think it needs to be any more complicated than just one for the addition of the feature to check these attributes. It could go in core or here. No opinion on where is better.

jvz:
The core PR might make more sense since you can only have one realm at a time anyways.

src/main/java/hudson/security/UserAttributesHelper.java
jeffret-b: (This is really a nit, but since we started discussing it, I'll continue a bit. It's not a blocker.)
The documentation indicates otherwise.

```
public static Boolean valueOf(String s)

Returns a Boolean with a value represented by the specified string. The Boolean returned represents a true value if the string argument is not null and is equal, ignoring case, to the string "true".
```

I guess yours results in a tri-valued Boolean (true, false, or null), whereas I suspect the standard one only returns (Boolean.FALSE, Boolean.TRUE). If you need that third value for some reason, then yours is better, otherwise it would be better to have a dual-valued boolean.

jeffret-b (approved):
Looks great.

I think we need to get an escape hatch in somewhere before introducing this capability, but that's not a blocker for this PR.

jglick:
Best to keep on hold until after #49.

Do we really need to take this approach, plus https://github.com/jenkinsci/jenkins/pull/4925? Seems like this could all be handled simply as an implementation improvement in the `ldap` plugin: if the account is “locked” or “expired” or whatever, throw an exception from `loadUserByUsername` rather than setting some obscure flags on the `UserDetails` and relying on a higher layer to check for those flags. Is there any actual use case for loading a `UserDetails` with some of these flags set?

jvz:
> Do we really need to take this approach, plus [jenkinsci/jenkins#4925](https://github.com/jenkinsci/jenkins/pull/4925)?

It certainly simplifies things when we go to patch active-directory-plugin to do the same thing. It's possible that other security realms can expose this metadata without authenticating the user. I filed https://issues.jenkins-ci.org/browse/JENKINS-63860 not that long ago as an alternative feature so that admins can simply clear out or delete any security data for disabled/expired users.

jglick:
> It certainly simplifies things when we go to patch active-directory-plugin to do the same thing.

How is anything simplified? If you need to patch `active-directory-plugin` then you could equally easily patch it throw authentication exceptions, right? Am I missing something?

jvz:
These attributes and exceptions are typically handled by Spring Security itself. Since we bypass the servlet filter and any calls to `AuthenticationManager.authenticate()` in API tokens and SSH keys (among any other uses of `User.impersonate()` which rely on the user being valid), it seemed the easiest way to address this was to add similar checks to what is done in Spring's abstract authentication manager class. Current version: https://github.com/spring-projects/spring-security/blob/master/core/src/main/java/org/springframework/security/authentication/AbstractUserDetailsReactiveAuthenticationManager.java (there's very similar code in Acegi Security, but not in the same place of course)

jglick:
> These attributes and exceptions are typically handled by Spring Security itself.

Right, I am just asking if it buys us anything concrete to set `accountNonExpired = false` and wait for Jenkins core to reject the authentication, rather than simply having the `SecurityRealm` throw the likes of `AccountExpiredException`—or some more generic supertype, it really does not matter.

My PoV from jenkinsci/jenkins#4848 is that we would prefer to minimize the scope of Spring Security APIs which are used by the plugin-core interface. So if we can keep it simple and say “this method can either return a valid `Authentication` or throw an exception”, that would seem less error-prone.

If there are some conditions under which implementation code _in Spring Security_ sets¹ these flags, then that would be a very different situation. But at least in the case of LDAP that does not seem to be the case—or if it were, then why would we need this patch at all?

----

¹Technically, “clears” or “resets” these flags. The negative sense in method names is confusing. Would prefer `isAccountExpired`.

jglick:
> or some more generic supertype, it really does not matter

…except to the extent that it creates extra work if I have to add new deprecated exception type mirrors to https://github.com/jenkinsci/jenkins/pull/4848, so more generic exception types already mirrored should be preferred.

jvz:
One possibility seems to be that the plugin's `UserDetailsService` can throw a `UsernameNotFoundException` (though not the more appropriate ones that get thrown by the user details validation class). Otherwise, the wrapping `ImpersonatingUserDetailsService` will simply catch the other exceptions and still attempt an impersonation using the last granted authorities (which would defeat the point of disabling an account remotely). If I change the `UserDetailsService` to throw the more appropriate security exceptions, then that would likely cause code to fail due to missed exceptions when almost all uses of `User.impersonate()` only try to catch `UsernameNotFoundException` and not its sibling exception types (likely due to the method signature and lack of any previous code throwing the other exceptions like we could do here).

jvz:
> If there are some conditions under which implementation code _in Spring Security_ sets¹ these flags, then that would be a very different situation. But at least in the case of LDAP that does not seem to be the case—or if it were, then why would we need this patch at all?

I checked spring-ldap, and they do not bother loading these attributes since they'd normally be checked at the LDAP server level when you try to bind to that user and password. That's how it already works in Jenkins, too, but Jenkins offers alternative authentication mechanisms that bypass that authentication flow. The only place I've ever seen these attributes filled in are the various JDBC or embedded user databases; I have yet to find an example in the wild of integrating external auth providers to this degree of granularity since it's usually the job of the auth provider to check account validity or other business logic behind authentication.

jglick:
> the wrapping `ImpersonatingUserDetailsService` will simply catch the other exceptions and still attempt an impersonation

Only if you throw `UserMayOrMayNotExistException` or `DataAccessException`. Throw some other type, like `UsernameNotFoundException`, and it will not.

I would advise putting both of these PRs on hold until after JEP-227 since the compatibility code for exception hierarchies is subtle and it would be easy to make a mistake. In Spring land, `ImpersonatingUserDetailsService2` checks only for `UserMayOrMayNotExistException2`, which subsumes the former functions of both `UserMayOrMayNotExistException` and `DataAccessException`.

src/main/java/hudson/security/UserAttributesHelper.java
```diff
@@ -0,0 +1,285 @@
...
+    private static final int ADS_UF_DISABLED = 0x0002;
+    private static final int ADS_UF_LOCK_OUT = 0x0010;
+    private static final int ADS_DONT_EXPIRE_PASSWORD = 0x1_0000;
+    private static final int ADS_UF_PASSWORD_EXPIRED = 0x80_0000;
+
+    private static final ChronoFormatter<Moment> GENERALIZED_TIME_FORMATTER;
+
+    static {
+        // https://stackoverflow.com/a/42383153
+        // this static block licensed https://creativecommons.org/licenses/by-sa/3.0/
```
jvz, line 90: Note to self: since this code snippet is licensed differently, it might be a good idea to split into its own file with the proper license header.

jvz:
Closing this to recreate against #49.

PR #59, opened 2020-11-13
[JENKINS-55813] Add support for LDAP password policy attributes (account locking)

jvz:
Builds on top of #49, ~so diff against Jesse's branch for an easier view: https://github.com/jglick/ldap-plugin/compare/spring-5303-new...jvz:JENKINS-55813-spring-security~

https://issues.jenkins.io/browse/JENKINS-55813

PR #60, opened 2020-11-24
Throw AccountStatusException rather than setting flags on LdapUserDetails

jglick:
Amends #59 to not rely on patches elsewhere. See https://github.com/jenkinsci/jenkins/pull/4925#issuecomment-733116395 for discussion.

### DIFF
--- a/src/main/java/hudson/security/UserAttributesHelper.java
+++ b/src/main/java/hudson/security/UserAttributesHelper.java
@@ -36,6 +36,10 @@ import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
+import org.springframework.security.authentication.AccountExpiredException;
+import org.springframework.security.authentication.CredentialsExpiredException;
+import org.springframework.security.authentication.DisabledException;
+import org.springframework.security.authentication.LockedException;
 
 /**
  * Provides helpers for checking operational attributes of LDAP user accounts for validating an account is still in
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/security/UserAttributesHelper.java
```diff
@@ -80,90 +84,118 @@
     private static final int ADS_UF_PASSWORD_EXPIRED = 0x80_0000;
 
     // https://ldapwiki.com/wiki/Administratively%20Disabled
-    public static boolean checkIfUserEnabled(@NonNull Attributes user) {
+    public static void checkIfUserEnabled(@NonNull Attributes user) throws DisabledException {
         // Active Directory attributes
         Integer uac = getUserAccountControl(user);
         if (uac != null && (uac & ADS_UF_DISABLED) == ADS_UF_DISABLED) {
-            return false;
+            throw new DisabledException("TODO details");
```
jglick, line 91: Besides avoiding the need to patch other components, a minor benefit is that we can specify diagnostic details right here in the stack trace rather than trying to synchronize log messages.

jvz:
Oooh, you've found a new feature from spring-security that would help a lot! I'll take a closer look in a bit.

src/test/java/hudson/security/LDAPEmbeddedTest.java
```diff
@@ -612,7 +614,10 @@ public void userValidityAttributes() throws Exception {
         LDAPSecurityRealm realm = new LDAPSecurityRealm(Collections.singletonList(configuration), false, null, null, null);
         r.jenkins.setSecurityRealm(realm);
         r.configRoundtrip();
-        assertFalse(realm.loadUserByUsername2("amy").isEnabled());
-        assertFalse(realm.loadUserByUsername2("bender").isAccountNonExpired());
+        assertThrows(DisabledException.class, () -> realm.loadUserByUsername2("amy"));
+        assertThrows(DisabledException.class, () -> User.getById("amy", true).impersonate2());
+        assertThrows(AccountExpiredException.class, () -> realm.loadUserByUsername2("bender"));
+        assertThrows(AccountExpiredException.class, () -> User.getById("bender", true).impersonate2());
+        assertThrows(FailingHttpStatusCodeException.class, () -> r.createWebClient().withBasicApiToken("amy").goTo(""));
```
jglick, line 621: Currently a 500, with the `DisabledException` being printed to the system log, pending some nicer handler along the lines of `AccessDeniedHandlerImpl`.

jvz (approved):
Oooh this is great. I love the stacktrace, too, because I had to add temporary logging statements otherwise while testing. I'll port over the message details from core.

jglick:
Just file a subsuming PR if you like (`merge`).

src/test/java/hudson/security/LDAPEmbeddedTest.java
jvz: Would that be a core change? Or can I specify an `AccessDeniedHandler` bean in this plugin?

jglick: Core change I guess. Not sure it is worth the effort anyway.

PR #61, opened 2020-11-24
Use more informative exceptions for user attribute validation

jvz:
This updates the feature introduced in [JENKINS-55813](https://issues.jenkins-ci.org/browse/JENKINS-55813) to use spring-security exceptions when loading user details rather than relying on downstream callers to validate the attributes. This PR subsumes #60.

@jglick @jeffret-b @Wadeck @daniel-beck


jglick (approved):

jeffret-b (approved):
This looks good. A nice, cleaner approach using the more informative exceptions.

I haven't tested any of the behaviors, but the code looks good and makes sense.

rsandell (approved):

PR #70, opened 2021-01-12
Use digest to get base ldap docker image

alecharp:
The docker image used to have a LDAP server for the tests was changed recently. And using `latest` couldn't warn us of that. However, using the digest, we will at least know when this image is not available anymore.

This is required for PCT run.


MRamonLeon (approved):
Fair enough.

MRamonLeon:
@rsandell could you take a look to this please?

rsandell (approved):

PR #81, opened 2021-04-01
[JENKINS-65117] fix the case sensitivity in the test method for the DN

jtnord:
[JENKINS-65117](https://issues.jenkins.io/browse/JENKINS-65117)

A DN is hierachical and contains attribute names and their values.  

An attribute name is case insensitive so when comparing a String representaion of a DN you will have incorrect results if one side has a different case for the attribute name than the other.

The solution is to parse the DN and compare the parsed results.

Manually tested with the planet express test DB and with a user that has some characters that need to be escaped as part of its CN (a comma).

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!


src/main/java/jenkins/security/plugins/ldap/LDAPExtendedTemplate.java
```diff
@@ -60,7 +60,7 @@ public LDAPExtendedTemplate(ContextSource dirContextFactory) {
      * @return the first matching entry converted using the specified {@link LdapEntryMapper}, or null if no matching
      * entry was found.
      */
-    public <T> @CheckForNull T searchForFirstEntry(@NonNull final String base, @NonNull final String filter,
+    public @CheckForNull <T> T searchForFirstEntry(@NonNull final String base, @NonNull final String filter,
```
jtnord, line 63: eclipse would not compile without this..
```
Unresolved compilation problem:
        The annotation @CheckForNull is disallowed for this location
```

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -7,7 +7,7 @@ LDAPSecurityRealm.AuthenticationSuccessful=Authentication: successful
 LDAPSecurityRealm.AuthenticationFailed=Authentication: failed for user "{0}"
 LDAPSecurityRealm.AuthenticationFailedEmptyPass=Authentication: failed for user "{0}" with empty password
 LDAPSecurityRealm.UserId=User ID: {0}
-LDAPSecurityRealm.UserDn=User Dn: {0}
+LDAPSecurityRealm.UserDn=User DN: {0}
```
jtnord, line 10: always `DN` never `Dn`

car-roll (approved):

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1778,11 +1779,25 @@ public FormValidation validate(LDAPSecurityRealm realm, String user, String pass
...
-                                    loginDetails.getDn(), lookUpDetails.getDn()));
-                    potentialLockout = true; // consistency is important
+                try {
+                    // do not use string comparison here as Attributes are always case insentive
+                    // the values may or may not be and that depend in part on the attributetype
+                    // and configuration
+                    LdapName loginDN = new LdapName(loginDetails.getDn());
+                    LdapName lookupDN = new LdapName(lookUpDetails.getDn());
+                    if (!loginDN.equals(lookupDN)) {
+                        // do not use the loginDN.toString() directory as this returns the unparsed version
```
jtnord, line 1789: I guess I should address this note to self

kerogers-cloudbees (approved):
Change looks good to me.

src/main/java/hudson/security/LDAPSecurityRealm.java
jtnord, line 1789:
```suggestion
                        // TODO do not use the loginDN.toString() directory as this returns the unparsed version
```

MRamonLeon (approved):

bitwiseman (approved):

rsandell (approved):

PR #82, opened 2021-04-01
Add CODEOWNERS

jtnord:
Add a codeowners file so that the maintainers get added as reviewers (normal people do not have access to do so)

- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,8 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @global-owner1 and @global-owner2 will be requested for
+# review when someone opens a pull request.
+* rsandell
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,8 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @global-owner1 and @global-owner2 will be requested for
+# review when someone opens a pull request.
+* rsandell
```
basil, line 8: I think following the example in [the archetype](https://github.com/jenkinsci/archetypes/blob/ff60037cf37185fc33f2270946a8dd8e907e75fa/common-files/.github/CODEOWNERS) would be easier to maintain, since it relies on the centrally maintained group name rather than on individual GitHub usernames:

```suggestion
* @jenkinsci/ldap-plugin-developers
```

PR #86, opened 2021-04-19
Add the choice of whether to perform certificate verification when connect to ldaps://

longping-t:
When connecting to LDAP server over SSL/TLS(LDAPS) with self-signed certificate, the jenkins server cannot connect to LDAP server.
The error log shows that [Root exception is javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target].

If we can get the cert from LDAP server, we import the cert to trust cacerts in jenkins server. Then the above error can be resolved. But if there are multiple servers, this is a bit troublesome.

Another way is to skip certificate validation.

Add the choice "SSL Verify" in LDAP configuarion for users to choose:

1. If checked and ldap server is an ldaps:// style URL, requiring the certificate to be verified.
2. If unchecked (the default), Jenkins will not verify the server certificate.


src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -250,6 +250,14 @@
     @Deprecated @Restricted(NoExternalUse.class)
     public transient String server;
 
+    /**
+     * whether to verify ldaps sever certificate? default is false
+     */
+    @SuppressFBWarnings(value = "UUF_UNUSED_PUBLIC_OR_PROTECTED_FIELD",
+            justification = "This public field is exposed to the plugin's API")
+    @Deprecated @Restricted(NoExternalUse.class)
+    public transient boolean sslVerify;
```
rsandell, line 259: You are adding a new deprecated, transient field, meaning it should not be used and will not be stored on disk.

I also don't think this should be set on the global level but on each separate configuration.
The default should also be `true` since that is the old behavior. And you need to make sure that old configurations are set to `true` when upgrading the plugin, so that Jenkins doesn't suddenly start behaving differently (less secure) in that sense.

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -415,75 +423,75 @@ group target (CN is a reasonable default)
      * @deprecated retained for backwards binary compatibility.
      */
     @Deprecated
-    public LDAPSecurityRealm(String server, String rootDN, String userSearchBase, String userSearch, String groupSearchBase, String managerDN, String managerPassword, boolean inhibitInferRootDN) {
-        this(server, rootDN, userSearchBase, userSearch, groupSearchBase, managerDN, managerPassword, inhibitInferRootDN, false);
+    public LDAPSecurityRealm(String server, boolean sslVerify, String rootDN, String userSearchBase, String userSearch, String groupSearchBase, String managerDN, String managerPassword, boolean inhibitInferRootDN) {
```
rsandell, line 426: these are old deprecated constructors and are kept here to keep backwards binary compatibility. So there signature should never change.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -141,13 +146,14 @@
     private transient String id;
 
     @DataBoundConstructor
-    public LDAPConfiguration(@NonNull String server, String rootDN, boolean inhibitInferRootDN, String managerDN, Secret managerPasswordSecret) {
+    public LDAPConfiguration(@NonNull String server, boolean sslVerify, String rootDN, boolean inhibitInferRootDN, String managerDN, Secret managerPasswordSecret) {
```
rsandell, line 149: The old constructor signature should be kept but deprecated, calling the new signature with the default value for the new field.

rsandell (changes requested):
Can you write some for of test for this?

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
longping-t: New commit resolved. Please help review.

src/main/java/hudson/security/LDAPSecurityRealm.java
longping-t: New commit resolved. Please help review.

src/main/java/hudson/security/LDAPSecurityRealm.java
longping-t: New commit resolved. Please help review.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -100,6 +100,11 @@
      */
     private final String server;
 
+    /**
+     * whether to verify ldaps sever certificate? default is false
+     */
+    private boolean sslVerify;
```
rsandell, line 106: This still defaults to `false` when it should default to `true`

Set the field to `true` in the `@DataboundConstructor`, the `@DataboundSetter`s will be called after the constructor and set the value to `false` if it was configured as such in the form field. old `JCasC` configs should then also still work.
You also need to set the default value in the form.

Next you also need to make sure that old configurations that are loaded from disk gets the default `true` value. The common way of solving that would be to make this field an object `Boolean` instead so that you can distinguish between a configured and non configured value. and then in the `readResolve` method set the field to `true` if it is `null`.
See https://wiki.jenkins.io/display/JENKINS/Hint+on+retaining+backward+compatibility for a bit more hints on how it works.

Then you should/could also add a test to the backwards compatibility. I am somewhat sure there already are some test for it so you can just add a `assertTrue(configuration.isSslVerify())` to it.

rsandell (changes requested):
Please avoid force pushing in the future if possible. It makes reviewing harder as I can't see what has changed since I last reviewed.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
longping-t: Because the @DataboundSetters will be called after the constructor, the constructor cannot get the value configured in the form field.
This field is referenced in inferRootDN() . And the constructor calls inferRootDN() to infer the root DN when rootDN is blank and it is not allowed to be blank.  
So  calling the new signature with the default value for the new field. The old constructor signature is kept but deprecated.
Please help review the new changes.

WilliamBehrens:
It would also be nice to add some information to the readme about enabling ldaps:// currently even using Google returns 0 results about how to configure it in Jenkins

Tethik:
>If unchecked (the default), Jenkins will not verify the server certificate.

It's bad security practice to opt out of certificate validation by default as it opens up for man-in-the-middle attacks. Please leave it on, and allow users to opt out if needed.

longping-t:
> > If unchecked (the default), Jenkins will not verify the server certificate.
> 
> It's bad security practice to opt out of certificate validation by default as it opens up for man-in-the-middle attacks. Please leave it on, and allow users to opt out if needed.

I have modified the default case to need to verify the certificate. Thank you for your advice.

jtnord:
I would say this is not something that should be added.
If there is a need to support a self signed cert in 2024 then is there a reason that this can not be added to the JVM keystore?

jtnord:
> If we can get the cert from LDAP server, we import the cert to trust cacerts in jenkins server. Then the above error can be resolved. But if there are multiple servers, this is a bit troublesome.

FWIW: you should have a common root cert for these servers and then you can just import that root cert, not all of the certs.

freafrea:
> I would say this is not something that should be added. If there is a need to support a self signed cert in 2024 then is there a reason that this cert can not be added to the JVM keystore?

In my case it is an self-signed cert that is incorrectly generated and java is unable to trust it. As this is a corporate environment I do not have any way to force a fixed cert. Unfortunately the only solution I found is just skipping checks for these certs.

PR #93, opened 2021-05-12
[JENKINS-65628] fix root DNs with spaces

jtnord:
[JENKINS-65628](https://issues.jenkins.io/browse/JENKINS-65628)  Root DNs containing spaces work again.

The switch to Spring Security broke the ability to authenticate users where a root DN contained a space (or any other character that needed URI escaping).

This works around https://github.com/spring-projects/spring-security/issues/9742 by providing our own `SpringSecurityContextSource` that fixes the munged DN.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

@rsandell @MRamonLeon @jglick


src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -186,7 +186,10 @@ public String getRootDN() {
     }
 
     public String getLDAPURL() {
-        return LDAPSecurityRealm.toProviderUrl(getServerUrl(), fixNull(rootDN));
+        
+        String s = LDAPSecurityRealm.toProviderUrl(getServerUrl(), fixNull(rootDN));
+        System.out.println("******* getLDAPURL -> " + s);
+        return s;
```
jtnord, line 192:
```suggestion
        return LDAPSecurityRealm.toProviderUrl(getServerUrl(), fixNull(rootDN));
```

src/test/java/jenkins/security/plugins/ldap/LDAPDNEscapingTest.java
```diff
@@ -0,0 +1,129 @@
...
+        return new Object[] {
+                  // DN, user search base, group search base
+                  new Object[] { null, null, null },
+                  new Object[] { "dc=planet express,dc=com", null, null },
+                  new Object[] { null, "dc=planet express,dc=com", null },
+                  new Object[] { null, null, "dc=planet express,dc=com" },
+                  new Object[] { "dc=com", "dc=planet express", "dc=planet express" }
+                  };
+    }
+}
```
jtnord, line 129:
```suggestion
}

```

src/test/java/jenkins/security/plugins/ldap/LDAPDNEscapingTest.java
```diff
@@ -0,0 +1,129 @@
...
+        try (InputStream ldifIs = LDAPDNEscapingTest.class.getResourceAsStream("/jenkins/security/plugins/ldap/LDAPDNEscapingTest/dnWithSpaces.ldif")) {
+            ads.loadSchema("planetexpress", "dc=planet express,dc=com", ldifIs);
+       }
+    }
+
+    // here as eclipse can not run a single parameterized test as it does not support the custom JUnitParamsRunner
+    //@Test
+    public void testSpacesInDN() throws Exception {
+        testOrgEscaping("dc=planet express,dc=com", null, null);
+    }
```
jglick, line 76:
```suggestion
    /*
    @Test
    public void testSpacesInDN() throws Exception {
        testOrgEscaping("dc=planet express,dc=com", null, null);
    }
    */
```

src/test/java/jenkins/security/plugins/ldap/LDAPDNEscapingTest.java
```diff
@@ -0,0 +1,129 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.hasSize;
+
+/**
+ * Tests {@link LDAPConfiguration} with DNs that are not URL safe.
+ */
+@LDAPTestConfiguration
+@RunWith(JUnitParamsRunner.class)
```
jglick, line 55: Why the special test dep, when JUnit has built-in support for parameterized tests?

jglick (approved):
I guess? Do not know what this really means.

src/test/java/jenkins/security/plugins/ldap/LDAPDNEscapingTest.java
jtnord: it is cleaner with this runner, you can have different tests with different parameters etc (which I had at one point) rather than littering a class with constructors or parameters.

mentioned in passing in the official junit docs for parameterized tests. https://github.com/junit-team/junit4/wiki/Parameterized-tests#see-also  so it is not that "far out"

src/test/java/jenkins/security/plugins/ldap/LDAPDNEscapingTest.java
jtnord: does this matter?  at least in eclipse uncommenting the test would be twice as much work :-p

jglick: Just for clarity, so that there is not an unused method lying around.

src/test/java/jenkins/security/plugins/ldap/LDAPDNEscapingTest.java
jglick: :shrug: just hoping to avoid a new dep.

Or you can just define a single plain test and run through these parameters in a loop, which would also be faster (there is no particular reason to start a new Jenkins service each time).

jglick (approved):

jtnord:
the reasons etc are mostly commented in the upstream ticket.

But in summary, 

SPring takes the LDAP url like `ldap://server/dc=foo%bar,dc=wibble` and extracts the `base DN` from it (using an incorrect approach to get `dc=foo%bar,dc=wibble` .  this then gets set as the `base` and then used in the call to the LDAP server.
THis fails obviously.  
The quick and dirty workaround fix (so we do not have to wait for a Spring security fix/release and core bump) is to trap the call to `setBase` and then attempt to decode the DN.   if we can decode it then we use it, if doceding fails because say it has already been decoded then it will attempt to use what was passed.

src/test/java/jenkins/security/plugins/ldap/LDAPDNEscapingTest.java
jtnord: I had various failures in the different branches so seeing a full list of the failures for the different parameters was helpful to me.

jglick (approved):

jtnord:
build of this has been used on the instance showing the issue and it works as expected.

src/main/java/jenkins/security/plugins/ldap/FixedDefaultSpringSecurityContextSource.java
```diff
@@ -0,0 +1,38 @@
+package jenkins.security.plugins.ldap;
+
+import java.net.URI;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.springframework.security.ldap.DefaultSpringSecurityContextSource;
+
+/*
```
rsandell, line 8:
```suggestion
/**
```

src/test/resources/jenkins/security/plugins/ldap/LDAPDNEscapingTest/dnWithSpaces.ldif
```diff
@@ -0,0 +1,2451 @@
...
+objectClass: organizationalPerson
+objectClass: person
+objectClass: top
+cn: Bender Bending Rodríguez
+sn: Bending Rodríguez
+description: Robot
+displayName: Bender
+employeeType: Ship's Robot
+givenName: Bender
+jpegPhoto:: /9j/4AAQSkZJRgABAQEAYABgAAD/4QBgRXhpZgAASUkqAAgAAAACADEBAgAHAAAA
```
rsandell, line 39: _nit_ You could have skipped these attributes to save some space

rsandell (approved):

PR #97, opened 2021-06-13
[NEEDS REVIEW] Fixes JENKINS-14520 - LDAP StartTLS support

AnalogJ:
Hi!
This PR fixes [JENKINS-14520](https://issues.jenkins.io/browse/JENKINS-14520). 
The Jenkins LDAP Plugin does not currently support LDAP over TLS (StartTLS), only the non-standardized (and deprecated) LDAPS.

In this PR, I've modified both the `javax.naming.ldap.LdapContext` and Spring-LDAP usage to support STARTTLS. **I don't have a LDAPS server configured to test if LDAPS support has been broken**

The changes are fairly minimal, and do not touch the authn/authz mechanisms at all, just the connection configuration. I've also packaged this code and it's confirmed working on a Jenkins server. 

I'm not really familiar with Spring (or Java development really), so an edit/review of the code would be appreciated (**especially the error handling and `.close()` functionality**). 

### Links
Most of my code came from research and examples that I found online, which I've compiled and collected here: https://github.com/AnalogJ/docker-jenkins-playground 

I used some of @mjalbanna code (https://github.com/jenkinsci/ldap-plugin/compare/master...mjalbanna:master) but his code was based on 1.x rather than 2.x, so it didn't have any spring-ldap fixes. 

### Tests
I have not written any tests, mostly because my thrown-together Java development environment doesn't seem to like the existing test suite. I had to package the plugin with `mvn package -Dmaven.test.skip=true`. I'm happy to add tests if someone can provide some guidance.

### DIFF
--- a/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
+++ b/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
@@ -54,6 +54,7 @@ import org.springframework.security.ldap.DefaultSpringSecurityContextSource;
 import org.springframework.security.ldap.search.FilterBasedLdapUserSearch;
 import org.springframework.security.ldap.search.LdapUserSearch;
 import org.springframework.security.ldap.userdetails.LdapAuthoritiesPopulator;
+import org.springframework.ldap.core.support.DefaultTlsDirContextAuthenticationStrategy;
 
 import javax.naming.Context;
 import javax.naming.NamingException;
@@ -614,6 +615,10 @@ public class LDAPConfiguration extends AbstractDescribableImpl<LDAPConfiguration
         vars.put("com.sun.jndi.ldap.read.timeout", "60000"); // timeout if no response after 60 seconds
         vars.putAll(getExtraEnvVars());
         contextSource.setBaseEnvironmentProperties(vars);
+
+        // pooled connections do not work with StartTLS
+        contextSource.setPooled(false);
+        contextSource.setAuthenticationStrategy(new DefaultTlsDirContextAuthenticationStrategy());
         contextSource.afterPropertiesSet();
 
         FilterBasedLdapUserSearch ldapUserSearch = new FilterBasedLdapUserSearch(getUserSearchBase(), getUserSearch(), contextSource);
--- a/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
+++ b/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
@@ -61,6 +61,12 @@ import javax.naming.directory.Attribute;
 import javax.naming.directory.Attributes;
 import javax.naming.directory.DirContext;
 import javax.naming.directory.InitialDirContext;
+import javax.naming.ldap.InitialLdapContext;
+import javax.naming.ldap.LdapContext;
+import javax.naming.ldap.StartTlsRequest;
+import javax.naming.ldap.StartTlsResponse;
+import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.net.ssl.SSLSession;
 
 import java.io.IOException;
 import java.net.InetAddress;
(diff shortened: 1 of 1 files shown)


digennarot (approved):

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -405,18 +412,37 @@ public FormValidation doCheckServer(@QueryParameter String value, @QueryParamete
...
-                if(managerDN!=null && managerDN.trim().length() > 0  && !"undefined".equals(managerDN)) {
-                    props.put(Context.SECURITY_PRINCIPAL,managerDN);
+                props.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.ldap.LdapCtxFactory");
+                props.put(Context.PROVIDER_URL, LDAPSecurityRealm.toProviderUrl(server,rootDN));
+
+                LdapContext ctx = new InitialLdapContext(props, null);
+                StartTlsResponse tls = (StartTlsResponse) ctx.extendedOperation(new StartTlsRequest());
+                SSLSession session = tls.negotiate();
+
+                if (!session.isValid()) {
```
jtnord, line 422: does this not break for anyone using non SSL LDAP? (in other words this should be available for users to opt into, or opt out of).
For the AD plugin we also need to close the `StartTlsResponse` https://github.com/jenkinsci/active-directory-plugin/blob/19e9fbcf0518537c6c4bd24fb1fd09901e7fd60e/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java#L627-L644

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -405,18 +412,37 @@ public FormValidation doCheckServer(@QueryParameter String value, @QueryParamete
...
+                    throw new IOException("Couldn't negotiate StartTls session, session is invalid");
+                }
+                ctx.addToEnvironment(Context.SECURITY_AUTHENTICATION, "simple");
```
jtnord, line 425: is this needed, it would appear to be an unrelated change and would preclude the use of any stronger sasl algorithms?

jtnord (commented):
it would appear that this will break users who do not use a TLS enabled server as it is not opt in.

I am also not following why you need the same property in both the `LdapContext` and the `props ` hashtable.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
olamy: agree definitely using startTLS must be optional and not per default.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -577,6 +615,10 @@ public ApplicationContext createApplicationContext(LDAPSecurityRealm realm) {
         vars.put("com.sun.jndi.ldap.read.timeout", "60000"); // timeout if no response after 60 seconds
         vars.putAll(getExtraEnvVars());
         contextSource.setBaseEnvironmentProperties(vars);
+
+        // pooled connections do not work with StartTLS
+        contextSource.setPooled(false);
```
olamy, line 620: interesting. so you mean `com.sun.jndi.ldap.connect.pool` doesn't work anymore? it creates connectivity issues or just ignored? If so there are few usage this property in the code.

olamy (changes requested):

AnalogJ:
hey @jtnord @olamy 
Thanks so much for the comments on the PR. 
I definitely agree, the PR is not ready for merging as-is. To be honest, I'm not a Java developer, I was just lucky enough to find some good examples & forked code that I could merge together to get a working STARTTLS build. 

I was hoping someone with a stronger Java background would be willing to pick up my crude implementation and finesse it into something that would be usable. Given the fact that LDAPS has been deprecated for a while, and there seems to be alot of people in the [community interested in this fix](https://issues.jenkins.io/browse/JENKINS-14520), I hope that's a reasonable ask.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
AnalogJ: agreed, this definitely breaks non-STARTTLS implementations. I commented below with some additional context for this PR.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
AnalogJ:
```
                ctx.addToEnvironment(Context.SECURITY_AUTHENTICATION, "simple");
```
is copy pasted from https://docs.oracle.com/javase/jndi/tutorial/ldap/ext/starttls.html#TLS%20with%20Simple%20Authentication

I did see the `External SASL Authentication` section in that link, however I wasn't familiar enough with SASL or the Jenkins UI to add additional form controls.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
AnalogJ: yeah, this one took me a while to figure out, but every example i was able to find on the internet related to StartTLS and Spring had `setPooled(false)`

> Explicitly disable connection pooling, or Spring may attempt to StartTLS twice on the same connection.

Also there seems to be some known issues as documented in the `spring-ldap` docs: https://docs.spring.io/spring-ldap/docs/1.3.2.RELEASE/reference/html/pooling.html

AnalogJ:
Any feedback/update on this PR?

WilliamBehrens:
Speaking of LDAP over SSL, that appears to currently be broken (in master that is) as there is no documentation to enable it and using an ldaps:// url causes an error in the configuration page

Tyrael:
the configuration page reporting error with the current plugin version is a separate issue, currently that always happen regardless if the settings are correct: https://issues.jenkins.io/browse/JENKINS-68748

PR #102, opened 2021-07-21
simplify connectivity check

olamy:
- for anonymous check we may have not the right to getAttributes so we must simply ensure we can connect
- timeout should be configurable, simplify some code as well

### DIFF
--- a/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
+++ b/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
@@ -61,10 +61,7 @@ import javax.naming.directory.Attribute;
 import javax.naming.directory.Attributes;
 import javax.naming.directory.DirContext;
 import javax.naming.directory.InitialDirContext;
-import javax.naming.ldap.Control;
 import javax.naming.ldap.InitialLdapContext;
-import javax.naming.ldap.LdapName;
-import javax.naming.spi.NamingManager;
 
 import java.io.IOException;
 import java.net.InetAddress;
@@ -88,7 +85,8 @@ import java.util.regex.Pattern;
 import static hudson.Util.fixEmpty;
 import static hudson.Util.fixEmptyAndTrim;
 import static hudson.Util.fixNull;
-import static org.apache.commons.lang.StringUtils.*;
+import static org.apache.commons.lang.StringUtils.isBlank;
+import static org.apache.commons.lang.StringUtils.isNotBlank;
 
 /**
  * A configuration for one ldap connection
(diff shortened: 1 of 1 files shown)


src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -573,8 +599,9 @@ public ApplicationContext createApplicationContext(LDAPSecurityRealm realm) {
         }
         contextSource.setReferral("follow");
         Map<String, Object> vars = new HashMap<>();
-        vars.put("com.sun.jndi.ldap.connect.timeout", "30000"); // timeout if no connection after 30 seconds
-        vars.put("com.sun.jndi.ldap.read.timeout", "60000"); // timeout if no response after 60 seconds
+        vars.put("com.sun.jndi.ldap.connect.pool", "true");
```
jtnord, line 602: Are we sure we want this, especially given it is not configurable?

jtnord: https://docs.spring.io/spring-ldap/docs/1.3.2.RELEASE/reference/html/pooling.html. Implies this poolingbis bad and there is no validation of the context.

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -403,19 +410,25 @@ public FormValidation doCheckServer(@QueryParameter String value, @QueryParamete
...
                 }
+                // normal
                 props.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.ldap.LdapCtxFactory");
                 props.put(Context.PROVIDER_URL, LDAPSecurityRealm.toProviderUrl(server,rootDN));
 
-                DirContext ctx = new InitialDirContext(props);
-                ctx.getAttributes("");
+                props.put("java.naming.referral", "follow");
+                props.put("com.sun.jndi.ldap.connect.timeout", Integer.toString(CONNECT_TIMEOUT));
+                props.put("com.sun.jndi.ldap.connect.pool", "true");
```
jtnord, line 428: Pooling a one of check makes no sense

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
olamy: the idea was to have the exact same code/configuration for this test and how connection are created when used for authz see https://github.com/jenkinsci/ldap-plugin/blob/992524769cc9616d69b035b247d21badd9baafc1/src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java#L602
it was totally different which is not better

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
olamy: answer here https://github.com/jenkinsci/ldap-plugin/pull/102#discussion_r705056385

src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
jtnord: AHH, that may explain some random hangs. Of. The plugin then that are fixed by restarting.
I suggest pooling the s disabled everywhere, or isboptbin

olamy: well sounds good. disabled per default but a configurable option?

olamy: well there is still an option to add env props

PR #170, opened 2022-08-05
Replace deprecated APIs and modernization

offa:
Replaces usages of deprecated APIs and some further modernizations.

--------------

### DIFF
--- a/src/main/java/hudson/security/LDAPSecurityRealm.java
+++ b/src/main/java/hudson/security/LDAPSecurityRealm.java
@@ -1473,6 +1473,7 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
         public static final String DEFAULT_MAILADDRESS_ATTRIBUTE_NAME = "mail";
         public static final String DEFAULT_USER_SEARCH = "uid={0}";
 
+        @NonNull
         public String getDisplayName() {
             return jenkins.security.plugins.ldap.Messages.LDAPSecurityRealm_DisplayName();
         }
@@ -2067,6 +2068,7 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
 
         @Extension public static class DescriptorImpl extends Descriptor<CacheConfiguration> {
 
+            @NonNull
             @Override public String getDisplayName() {
                 return "";
             }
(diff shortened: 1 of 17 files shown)


offa:
CI build fails due to:

> java.lang.UnsupportedClassVersionError: com/trilead/ssh2/ServerHostKeyVerifier has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -1526,12 +1529,12 @@ public boolean hasEnableSecurityForm() {
...
         @RequirePOST
         public FormValidation doValidate(StaplerRequest req) throws Exception {
-            if (!Jenkins.getActiveInstance().hasPermission(Jenkins.ADMINISTER)) {
+            if (!Jenkins.get().hasPermission(Jenkins.ADMINISTER)) {
                 // require admin to test
                 return FormValidation.ok();
             }
             // extract the submitted details
-            JSONObject json = JSONObject.fromObject(IOUtils.toString(req.getInputStream()));
+            JSONObject json = JSONObject.fromObject(IOUtils.toString(req.getInputStream(), Charset.defaultCharset()));
```
basil, line 1537: I realize this was just preserving existing behavior, so specific action item here, but I suspect the original behavior was wrong. In the few cases in core where we use `IUtils#toString` with a `StaplerRequset` object, we always pass in either `req.getCharacterEncoding()` or UTF-8 as the encoding.

offa: Is it safe to assume UTF-8 or is `getCharacterEncoding()` the better choice?

basil: I think `ServletRequest#getCharacterEncoding` is technically the more correct option since the HTTP request could come in via any encoding in theory, but I think the HTTP requests are usually (always?) UTF-8 in practice.

offa: Thanks! This a good chance for improvement; I'm going to submit a PR.

offa: --> #174

PR #174, opened 2022-08-18
Use encoding of the request

offa:
_See #170 discussion._

/cc @basil 

------------------

### DIFF
--- a/src/main/java/hudson/security/LDAPSecurityRealm.java
+++ b/src/main/java/hudson/security/LDAPSecurityRealm.java
@@ -1534,7 +1534,7 @@ public class LDAPSecurityRealm extends AbstractPasswordBasedSecurityRealm {
                 return FormValidation.ok();
             }
             // extract the submitted details
-            JSONObject json = JSONObject.fromObject(IOUtils.toString(req.getInputStream(), Charset.defaultCharset()));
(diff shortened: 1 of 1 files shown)


basil (approved):
Definitely a more logical choice than the JVM character encoding, which is completely unrelated to the encoding of the HTTP request. Would be great if you could set a breakpoint and verify that it is always UTF-8 in practice, since that is what we expect. (And if it isn't, then we clearly haven't understood the problem!) But that is optional since I think the risk of regression is low.

offa:
There's just one corner case: `getCharacterEncoding()` could theoretically return `null` in case the request hasn't specified an encoding. This would end in `IOUtils.toString()` using default encoding (again). Should we use UTF-8 as a fallback here instead?

> Would be great if you could set a breakpoint and verify that it is always UTF-8 in practice, since that is what we expect.

All right, I'll do some testing.

basil:
> Should we use UTF-8 as a fallback here instead?

I think that makes sense. Certainly it seems like a better fallback than the JVM's default encoding, which as I mentioned earlier is completely unrelated.

offa:
As expected the request was UTF-8 in my tests.

However, I have only Linux hosted systems around which use UTF-8 by default (`file.encoding`).

basil (approved):
Thanks for the PR!

jtnord (approved):

car-roll (approved):

PR #185, opened 2022-10-10
[JENKINS-68748] Adapt Test LDAP functionality to 2.346.x and later

Dohbedoh:
[JENKINS-68748](https://issues.jenkins.io/browse/JENKINS-68748):

* adapt the functionality to 2.346.x `realm` -> `securityRealm` https://github.com/jenkinsci/jenkins/pull/5417
* adapt the functionality to 2.361.x changes in validation error area https://github.com/jenkinsci/jenkins/pull/6460
* remove obsolete conditionals and methods that are not relevant post 2.212 (pom is past this core version already).
* add TODOs to remind ourselves to remove 2.346.x specific compatibility when core 2.361 or later is required.


Dohbedoh:
Still need to test against 2.361.1 and later as well.

Dohbedoh:
Works on 2.346.1 but some UI issues displaying the error message in 2.361.2. (only tested those 2 versions):

applyErrorMessage

Need to spend some more time on this

Dohbedoh:
Tested with 2.346.1:



Tested with 2.361.1:



Tested with current weekly 2.372:

src/main/resources/jenkins/security/plugins/ldap/validation/validate.jelly
```diff
@@ -78,6 +78,7 @@
       <img src="${imagesURL}/spinner.gif"/>
       ${attrs.progress}
     </div>
+    <!-- TODO: remove once pom is based off 2.361-->
```
jglick, line 81: IMO it is better to just change the `jenkins.version` to 2.361.x and delete the tech debt.

If you want to offer a fix to 2.346.x users, do it on a branch.

jglick (commented):
Does any of this have test coverage? The JS hacks are completely unsupported and liable to break at any time.

Dohbedoh:
Added a test for the UI. tested with 2.346.1 and 2.361.1.

src/main/resources/jenkins/security/plugins/ldap/validation/validate.jelly
Dohbedoh: I don't have rights to create a branch. @olamy @jtnord WDYT ?

olamy: @Dohbedoh  if you don't have write access no you cannot. I can do it for you. (It will cost you few croissants to deliver to my place 😜)
what branch name do you want?

olamy: done https://github.com/jenkinsci/ldap-plugin/tree/2.12.x

Dohbedoh: Thanks!

jtnord (commented):
the screenshot for the weekly looks a bit off - like the validate buttons are all disabled?

jtnord:
The functionality does not work for me.

checked out the PR
`mvn hpi:run`

go to configure global security.
select "LDAP" as the security realm
click "Test LDAP Settings"

expected result

lots of errors as there is no LDAP server defined.  y machine is not part of a domain..,

Additionally looking in the chrome dev console - no HTTP request is made over the network.

when entering a server url (pick something) then there is a call to checkServer - but still the Test LDAP Settings button does nothing.

jtnord (changes requested):
javascript error at runtime and does not work for me.

Dohbedoh:
@jtnord I noticed something... I could not reproduce what you explained. Test test buttong worked for me. And at some point nothing was working anymore. What I found out is that I was at `/configureSecurity/` instead of `/manage/configureSecurity/`. The latter works and is the one that we are redirected to when navigating through the UI. The first endpoint seems to be the "legacy" one and from where things fail...

Dohbedoh:
While debugging, found out that the `Behaviour.specify` is never called in a scenario where you have something else than LDAP Security selected in LDAP when you enter the Configure Global Security screen, and you then select LDAP and try to use the test button. Debugging the `validate.js` I found out that the `Behaviour.specify` was not called, as if the script was not loaded or something... (or the registration was never detected).
I eventually removed the hack in place and just used the `st:adjunct` instead.  And that fixed this problem. I retested different scenarios and it seems stable now, the test button works in all scenarios that were sometimes failing and the CSS is still properly applied.

cc @jtnord @jglick

Dohbedoh:
I see, I can't change the Jenkinsfile :)

jtnord:
> I see, I can't change the Jenkinsfile :)

If you extract the `Jenkinsfile` change to a new PR that can be merge separately (assuming the existing code builds fine on java11 and 17)

jglick:
I have write access. I will get the toolchain bump filed and ready.

jglick (approved):
Not qualified to comment on the JavaScript, but looks OK.

Dohbedoh:
ping @jtnord

olamy (approved):
even with a new UI test! what can be wrong ;)

jtnord:
it works but every time I do a new "test" I get another "error" rather than it replacing the old result.

The same applied in the main server entry too - is the JS trickery at fault or is this a jenkins core issue?

image

jglick:
> every time I do a new "test" I get another "error" rather than it replacing the old result

I think I have seen the same in other places like GitHub validation. Could be a core UI regression.

jglick:
> appears to be constrained to this plugin?

No, it seems to be a core regression: https://github.com/jenkinsci/jenkins/pull/6460#issuecomment-1466841649

jtnord (approved):

PR #194, opened 2022-11-08
[JENKINS-70035] Use User.getById() instead of User.get()

Kevin-CB:
You can find more information in [JENKINS-70035](https://issues.jenkins.io/browse/JENKINS-70035)

I had to adapt [LdapMultiEmbeddedTest#when_first_is_unavailable_and_ignorable_and_login_on_second_then_login](https://github.com/jenkinsci/ldap-plugin/blob/e5925c2efe39c4181b7ac77d2ebbb4f78ac7a856/src/test/java/hudson/security/LdapMultiEmbeddedTest.java#L347) & [LdapMultiEmbeddedTest#when_first_is_wrong_and_ignorable_and_login_on_second_then_login](https://github.com/jenkinsci/ldap-plugin/blob/e5925c2efe39c4181b7ac77d2ebbb4f78ac7a856/src/test/java/hudson/security/LdapMultiEmbeddedTest.java#L336) to match `AuthenticationServiceException` and not `UserMayOrMayNotExistException2` because they were not thrown anymore. 

As I did not fully understand their uses, I would ask you to review this part carefully to make sure that I didn't broke them.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue


src/test/java/hudson/security/LdapMultiEmbeddedTest.java
```diff
@@ -340,7 +340,7 @@ public void when_first_is_wrong_and_ignorable_and_login_on_second_then_login() t
                 allOf(containsString(FAILED_COMMUNICATION_WITH_LDAP_SERVER),
                         containsString(WILL_TRY_NEXT_CONFIGURATION),
                         containsString(planetExpress.getUrl())),
-                CoreMatchers.<Throwable>instanceOf(UserMayOrMayNotExistException2.class)));
+                CoreMatchers.<Throwable>instanceOf(AuthenticationServiceException.class)));
```
dwnusbaum, line 343: Sorry but I do not remember if the specific exception type here is important. What is the full stack trace in the logs before and after the change? Maybe the only difference is that there is a new outer exception?

jtnord: `UserMayOrMayNotExistException2` seems important as it is used to say "I can not tell you that this users exists. or not"  and as such can allow the use of a token (wheras another excpetion would be treated as the form - this user does not exist).

At the very least I think some manual testing with user impersonation needs to occur when the server is down (before and after this change)

PR #296, opened 2024-09-17
Do not allows users to authenticate with short passwords in fips mode

BorisYaoA:
### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -320,6 +320,16 @@
   </dependencyManagement>
 
   <build>
+    <!-- TODO until https://github.com/jenkinsci/maven-hpi-plugin/pull/652 is released -->
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.jenkins-ci.tools</groupId>
+          <artifactId>maven-hpi-plugin</artifactId>
+          <version>3.58-rc1621.0cb_cd49b_b_c68</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
     <plugins>
       <plugin>
         <groupId>org.jenkins-ci.tools</groupId>


src/main/java/jenkins/security/plugins/ldap/LDAPConfiguration.java
```diff
@@ -411,6 +413,9 @@ public FormValidation doCheckServer(@QueryParameter String value, @QueryParamete
             if(!Jenkins.get().hasPermission(Jenkins.ADMINISTER))
                 return FormValidation.ok();
 
+            if(FIPS140.useCompliantAlgorithms() && managerPassword.length() < 14)
```
BorisYaoA, line 416: Checking that we are running Jenkins in FIPS mode and that the password is FIPS compliant (see https://github.com/jenkinsci/jep/tree/master/jep/237 for Jenkins FIPS support)

src/test/java/jenkins/security/plugins/ldap/LDAPConfigurationWithFIPSTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jvnet.hudson.test.FlagRule;
+import org.jvnet.hudson.test.JenkinsRule;
+
+/**
+ * Tests {@link LDAPConfiguration} in FIPS mode.
+ */
+public class LDAPConfigurationWithFIPSTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
```
alecharp, line 19: Could you use a `RealJenkinsRule` so that we could benefit from https://github.com/jenkinsci/jenkins-test-harness/pull/829?

fcojfernandez (changes requested):
While the FormValidation is the way to inform the user, it won't prevent the user to click on Save button and store a non-compliant configuration. Also, with CasC you could save a unsecure configuration, so the proper fix would be:

* Add the FormValidation as you did so in the UI you can see there's an error
* Add a check to prevent the non-compliant configuration to be saved. You can do that either in the DataBoundConstructor, in the DataBoundSetter or in a `PostConstruct` method. The usual way is throwing an IllegalArgumentException like in https://github.com/jenkinsci/active-directory-plugin/blob/ff7b23e0c59f21c35486a5f8fb533097e43a7e81/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java#L283-L286, or a FormException and take advantage of https://github.com/jenkinsci/jenkins/pull/9495

rsandell (changes requested):
The ticket is about **user** passwords not being less than 14 characters. The manager password being too short is another ticket.

fcojfernandez:
Bobby is right, this PR is fixing another different issue. The fix for this ticket should be in https://github.com/jenkinsci/ldap-plugin/blob/d3700c27b0435bcfcf4f0245864c312ed7988707/src/main/java/hudson/security/LDAPSecurityRealm.java#L241

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -753,6 +754,9 @@ public SecurityComponents createSecurityComponents() {
      */
     @Override
     protected UserDetails authenticate2(String username, String password) throws AuthenticationException {
+        if(FIPS140.useCompliantAlgorithms() && StringUtils.isNotBlank(password) && password.length() < 14) {
```
rsandell, line 757: I believe a blank password is also not allowed.

```suggestion
        if(FIPS140.useCompliantAlgorithms() && (StringUtils.isBlank(password) || password.length() < 14)) {
```

src/main/java/hudson/security/LDAPSecurityRealm.java
```diff
@@ -753,6 +754,9 @@ public SecurityComponents createSecurityComponents() {
      */
     @Override
     protected UserDetails authenticate2(String username, String password) throws AuthenticationException {
+        if(FIPS140.useCompliantAlgorithms() && (StringUtils.isBlank(password) || password.length() < 14)) {
+            throw new org.springframework.ldap.AuthenticationException(new javax.naming.AuthenticationException("When running in FIPS compliance mode, the password must be at least 14 characters long"));
```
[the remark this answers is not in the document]
BorisYaoA: I'm having this message `'AuthenticationException' is abstract; cannot be instantiated`

src/main/resources/jenkins/security/plugins/ldap/Messages.properties
```diff
@@ -6,6 +6,7 @@ LDAPSecurityRealm.LoginHeader=Login
 LDAPSecurityRealm.AuthenticationSuccessful=Authentication: successful
 LDAPSecurityRealm.AuthenticationFailed=Authentication: failed for user "{0}"
 LDAPSecurityRealm.AuthenticationFailedEmptyPass=Authentication: failed for user "{0}" with empty password
+LDAPSecurityRealm.AuthenticationFailedNotFipsCompliantPass=When running in FIPS compliance mode, the password must be at least 14 characters long
```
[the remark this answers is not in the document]
BorisYaoA: No I will remove it

fcojfernandez (approved):

olamy (approved):

PereBueno (approved):

rsandell (approved):

PR #299, opened 2024-09-25
[JENKINS-73825] ldap allows insecure configurations

vwagh-dev:
See [JENKINS-73825](https://issues.jenkins.io/browse/JENKINS-73825)
In FIPS mode, now LDAP don't allows the insecure configuration. Also it shows the error messages in case of insecure ldap url/ short password

### Testing
Manually testing is done with below scenarios
1. Configure insecure ldap url & error should thrown + form data should not be saved
2. Configure secure ldap url + form data should not be saved 
3. In advance configuration manager password less than 14 is not allowed
4. CASC configuration unit test to load the ldap configuration i.e. **CasCFipsTest**
     * When configuration has secure url + valid password length => Tested
     * When configuration has insecure url/valid password length **CasCFipsTest.configure_ldap_for_invalid** => Blocked
5. ConfigRoundTrip scenario with FIPS enabled i.e. **LDAPEmbeddedFIPSTest.login()** => Blocked

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!

### DIFF
--- a/src/test/java/hudson/security/LDAPEmbeddedFIPSTest.java
+++ b/src/test/java/hudson/security/LDAPEmbeddedFIPSTest.java
@@ -24,6 +24,7 @@
 
 package hudson.security;
 
+import hudson.util.FormValidation;
 import io.jenkins.plugins.casc.misc.ConfiguredWithCode;
 import jenkins.model.Jenkins;
 import hudson.tasks.MailAddressResolver;
@@ -35,6 +36,7 @@ import jenkins.model.IdStrategy;
 import jenkins.security.plugins.ldap.*;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
+import org.junit.rules.ExpectedException;
 import org.jvnet.hudson.test.FlagRule;
 import org.springframework.security.ldap.userdetails.LdapUserDetails;
 import org.junit.Rule;
@@ -65,6 +67,8 @@ public class LDAPEmbeddedFIPSTest {
     public RuleChain chain = RuleChain.outerRule(ads).around(r);
     @Rule
     public LoggerRule log = new LoggerRule();
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
 
     @ClassRule
     public static FlagRule<String> fipsSystemPropertyRule =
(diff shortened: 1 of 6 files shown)


src/test/java/hudson/security/LDAPEmbeddedFIPSTest.java
```diff
@@ -0,0 +1,158 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+
+@LDAPTestConfiguration(ldapsProtocol = true)
+public class LDAPEmbeddedFIPSTest {
+    public LDAPRule ads = new LDAPRule();
```
rsandell, line 64: I doubt that you need an actual ldap server running in the background for the tests you are going to do, it will only take up unnecessary compute.

src/test/java/hudson/security/LDAPEmbeddedFIPSTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+
+    @BeforeClass
+    public static void setUp(){
+        disableSSLVerification();
+    }
+
+    //@Test
+    @LDAPSchema(ldif = "planetexpress", id = "planetexpress", dn = "dc=planetexpress,dc=com")
+    public void login() throws Exception {
```
rsandell, line 85: none of the testing here looks relevant

src/test/java/hudson/security/LDAPEmbeddedFIPSTest.java
```diff
@@ -0,0 +1,158 @@
...
+       // disableSSLVerification();
+        System.out.println(ads.getUrl());;
+        String secureUrl = "ldaps://insecure.example.com";
+        LDAPSecurityRealm realm =
+                new LDAPSecurityRealm(ads.getUrlTls(), "dc=planetexpress,dc=com", null, null, null, null, null,
+                        "uid=admin,ou=system", Secret.fromString("pass"), false, false, null,
+                        null, "cn", "mail", null, null);
+        r.jenkins.setSecurityRealm(realm);
+        r.configRoundtrip();
+        String content = r.createWebClient().login("fry", "fry").goTo("whoAmI").getBody().getTextContent();
```
rsandell, line 95: This should fail when @BorisYaoA 's fix is merged as the password is too short.

src/test/java/hudson/security/LDAPEmbeddedFIPSTest.java
vwagh-dev: I'm just trying to login using the secure ldap url .

rsandell:
And don't reference Cb internal tickets in the PR description/title. Create a cloned ticket on issues.jenkins.io and reference that instead.

